PR #43, opened 2017-07-31
Using valid azure sdk arm version

dansanabria:
Pinning Gem versions to use v0.10.0 of azure sdk ARM until inspec azure
libraries are refactored to use newer methods introduced in v0.11.0 (WIP)
i.e. list_resources_as_lazy seems to be deprecated in v0.11.0

Signed-off-by: sanabriad <sanabria.d@gmail.com>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,9 +1,9 @@
 source 'https://rubygems.org'
 
 gem 'inifile'
-gem 'azure_mgmt_compute'
-gem 'azure_mgmt_resources'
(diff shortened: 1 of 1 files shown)


chris-rock:
@dansanabria I think we do need this anymore since #44 is merged. Am I correct?

dansanabria:
hey @chris-rock thanks , you are right this is not needed anymore.

chris-rock:
Awesome. Thank you @dansanabria

PR #55, opened 2017-10-26
Pinning version of Azure SDKs to avoid beaking change issue

### LINKED ISSUE #54: Peg Azure SDKs to version 0.14
Microsoft have added some breaking changes to the Azure SDKs 0.15 which means this plugin will not work with that version.

As a quick workaround the SDKs should be pinned to version 0.14 to allow it to keep working in the short term.

russellseymour:
Fixes #54

Signed-off-by: Russell Seymour <russell.seymour@turtlesystems.co.uk>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,9 +1,9 @@
 source 'https://rubygems.org'
 
 gem 'inifile'
-gem 'azure_mgmt_compute', '~> 0.5', '= 0.14.0'
-gem 'azure_mgmt_resources', '~> 0.5', '= 0.14.0'
-gem 'azure_mgmt_network', '~> 0.5', '= 0.14.0'
+gem 'azure_mgmt_compute', '= 0.14.0'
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -1,9 +1,9 @@
 source 'https://rubygems.org'
 
 gem 'inifile'
-gem 'azure_mgmt_compute', '~> 0.11'
-gem 'azure_mgmt_resources', '~> 0.11'
-gem 'azure_mgmt_network', '~> 0.11'
+gem 'azure_mgmt_compute', '~> 0.5', '= 0.14.0'
```
adamleff: If you have an equality pin, the pessimistic pin `~> 0.5` is unnecessary. I'd remove these to be abundantly clear what you're looking for.

russellseymour:
Issue #56 has been created to ensure that the plugin is updated to handle the new version of the Azure SDK

PR #63, opened 2018-06-29
Migrates to new version of resource pack

dmccown:
This new version is backed by a REST implementation. We are introducing
resources created for the specific use cases we are trying to solve, as
well as establishing a pattern that authors may follow to add new
resources.

The maintainers list will be updated in the near future with the correct
long term maintainers.


docs/resources/azurerm_virtual_machines.md.erb
```diff
@@ -0,0 +1,120 @@
...
+    # Insist that you have at least one virtual machine that starts with 'MyVm'
+    describe azurerm_virtual_machines(resource_group: 'MyResourceGroup').where { name.start_with?('WindowsVm') } do
+      it { should exist }
+    end
+
+## Properties
+
+* `os_disks`, `data_disks`, `vm_names`
+
+### os_disks
```
jerryaldrichiii: Should be `os\_disks`

docs/resources/azurerm_virtual_machines.md.erb
```diff
@@ -0,0 +1,120 @@
...
+
+Gives a list of OS disk names for all the virtual machines in the resource group.
+
+    its('os_disks.sort') { should eq ['MyDisk1', 'MyDisk2'] }
```
jerryaldrichiii: Could we give examples that avoid `.sort`, like:

```
its('os_disks') { should include 'DataDisk1' }
its('os_disks') { should include 'DataDisk2' }
```

docs/resources/azurerm_virtual_machines.md.erb
```diff
@@ -0,0 +1,120 @@
...
+
+### data_disks
```
jerryaldrichiii: Should be `data\_disks`

docs/resources/azurerm_virtual_machines.md.erb
```diff
@@ -0,0 +1,120 @@
...
+
+Gives a list of data disks for all the virtual machines in the resource group.
+
+    its('os_disks.sort') { should eq [['MyDisk1'], ['MyDisk2']] }
```
jerryaldrichiii: Could we give examples that avoid `.sort`, like:

```
its('os_disks') { should include 'DataDisk1' }
its('os_disks') { should include 'DataDisk2' }
```

docs/resources/azurerm_virtual_machines.md.erb
```diff
@@ -0,0 +1,120 @@
...
+
+Note that it returns an array of arrays. You may flatten them for testing:
+
+    its('os_disks.flatten.sort') { should eq ['MyDisk1', 'MyDisk2'] }
```
jerryaldrichiii: Could we give examples that avoid `.sort`, like:

```
its('os_disks') { should include 'DataDisk1' }
its('os_disks') { should include 'DataDisk2' }
```

docs/resources/azurerm_virtual_machines.md.erb
```diff
@@ -0,0 +1,120 @@
...
+
+### vm_names
```
jerryaldrichiii: Should be `vm\_names`

docs/resources/azurerm_virtual_machines.md.erb
```diff
@@ -0,0 +1,120 @@
...
+
+Gives a list of all the virtual machine names in the resource group.
+
+    its('vm_names') { should include('MyVm1', 'MyVm2') }
+
+## Matchers
+
+This InSpec audit resource has the following special matchers. For a full list
+of available matchers, please visit our [Universal Matchers
```
jerryaldrichiii: Newline in link

docs/resources/azurerm_virtual_machines.md.erb
```diff
@@ -0,0 +1,120 @@
...
+    end
+
+    # Should exist if the filter returns a single virtual machine
+    describe azurerm_virtual_machines(resource_group: 'MyResourceGroup').where('platform' => 'windows') do
+      it { should exist }
+    end
+
+## Azure Permissions
+
+Your [Service
```
jerryaldrichiii: New line in link

libraries/azure_monitor_log_profile.rb
```diff
@@ -0,0 +1,52 @@
...
+    retention_enabled: 'retentionPolicyEnabled',
+  }.freeze
+
+  attr_reader(*ATTRS.keys)
+
+  def initialize(options = { name: 'default' })
+    resp = client.log_profile(options[:name])
+    return if resp.nil? || resp.key?('error')
+
+    @name = resp['name']
```
jerryaldrichiii: Should we align the equals with the below?

libraries/azure_monitor_log_profile.rb
```diff
@@ -0,0 +1,52 @@
...
+      next if instance_variable_defined?("@#{name}")
+
+      instance_variable_set("@#{name}", fields[api_name])
+    end
+
+    @exists = true
+  end
+
+  def has_log_retention_enabled?
+    !!retention_enabled
```
jerryaldrichiii: Double not?

libraries/azure_network_security_group.rb
```diff
@@ -0,0 +1,108 @@
...
+  def destination_port_ranges(properties)
+    return Array(properties['destinationPortRange']) if properties['destinationPortRanges'].empty?
+    return properties['destinationPortRanges'] if properties['destinationPortRange'].empty?
+    properties['destinationPortRanges'].push(properties['destinationPortRange'])
+  end
+
+  def matches_port?(ports, match_port)
+    return true if ports.detect { |p| p =~ /^(#{match_port}|\*)$/ }
+    ports
+      .select { |port| port.include?('-') }
```
jerryaldrichiii: Usually I see the second method on the new line. Not sure if it matters though.

libraries/azure_security_center_policy.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+require 'azurerm_resource'
+
+class AzureSecurityCenterPolicy < AzurermResource
+  name 'azure_security_center_policy'
+  desc 'Verifies settings for Security Center'
+  example <<-EXAMPLE
```
jerryaldrichiii: I notice we use both `<<-EXAMPLE` and `"` should we standardize?

libraries/azurerm_virtual_machine.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+# TODO: we will have a name collision with the existing implementation
```
jerryaldrichiii: Could we reword this to be something we can do? Some action item?

libraries/azurerm_virtual_machine_disk.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+# TODO: we will have a name collision with the existing implementation
```
jerryaldrichiii: Could we reword this as something we can do?

libraries/azurerm_virtual_machine_disk.rb
```diff
@@ -0,0 +1,48 @@
...
+    @exists = true
+  end
+
+  def to_s
+    "'#{name}' Disk"
+  end
+
+  def encryption_enabled
+    return nil unless properties
+    !!properties.dig('encryptionSettings', 'enabled')
```
jerryaldrichiii: Double not?

libraries/support/azure/management.rb
```diff
@@ -0,0 +1,89 @@
...
+    def get(url:, api_version:)
+      confirm_configured!
+
+      cache.fetch(url) do
+        body = azure_client.get(url, params: { 'api-version' => api_version }).body
+        body.fetch('value', body)
+      end
+    end
+
+    def confirm_configured!
```
jerryaldrichiii: I thought `!` methods were only labeled as such if they modified the variable passed in or modified some external state.

test/functional/verify/controls/azure_network_security_group.rb
```diff
@@ -0,0 +1,24 @@
+resource_group = attribute('resource_group',         default: nil)
+nsg            = attribute('network_security_group', default: nil)
```
jerryaldrichiii: Are we aligning the `default:nil`?

Rakefile
```diff
@@ -1,101 +1,173 @@
...
 
-    # Create the plan that can be applied to Azure
-    cmd = format("cd %s/build/ && terraform plan -var 'subscription_id=%s' -var 'client_id=%s' -var 'client_secret=%s' -var 'tenant_id=%s' -var 'storage_account_name=%s' -var 'admin_password=%s' -var 'suffix=%s' -out inspec-azure.plan", integration_dir, creds[:subscription_id], creds[:client_id], creds[:client_secret], creds[:tenant_id], sa_name, admin_password, suffix)
-    sh(cmd)
+    puts stdout
 
-    # Apply the plan on Azure
-    cmd = format("cd %s/build/ && terraform apply inspec-azure.plan", integration_dir)
-    sh(cmd)
+    # TODO: add "warnings" back to list when we switch to using InSpec backend
```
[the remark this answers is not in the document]
dmccown: We can remove.

docs/resources/azure_resource_groups.md.erb
```diff
@@ -0,0 +1,95 @@
...
+The control will pass if the filter returns at least one result. Use
+`should_not` if you expect zero matches.
+
+    describe azure_resource_groups
+      it { should exist }
+    end
+
+## Azure Permissions
+
+Your [Service
```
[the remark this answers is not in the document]
dmccown: It's because the next line is that super long URI, that's how the line break function in vim broke it up. I can probably make it cleaner.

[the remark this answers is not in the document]
dmccown: hmm maybe let's old off and see if it becomes an issue with the document builder.

docs/resources/azure_security_center_policies.md.erb
```diff
@@ -0,0 +1,83 @@
...
+properties of some or all Azure Security Center Policies.
+
+Security Center Policies are defined for each Resource Group. A Security Center
+Policy called `default` also exists for every subscription.
+
+<br />
+
+## Syntax
+
+An `azure_security_center_policies` resource block uses an optional filter to
```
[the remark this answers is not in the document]
dmccown: We have example on line 36. Do you think it's worth repeating?

docs/resources/azure_security_center_policies.md.erb
```diff
@@ -0,0 +1,83 @@
...
+
+## Examples
+
+### Check for a Security Center Policy
+
+    describe azure_security_center_policies do
+      its('names') { should incldue 'default' }
+    end
+
+### Insist default Security Center Policy exists
```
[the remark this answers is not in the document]
dmccown: Assert instead of Insist?

docs/resources/azure_security_center_policy.md.erb
```diff
@@ -0,0 +1,206 @@
...
+
+<br />
+
+## Syntax
+
+An `azure_security_center_policy` resource block identifies a Security Center
+Policy by name. In the current Rest API you may only lookup a `default` policy.
+If no policy is given the default one will be used.
+
+    describe azure_security_center_policy(name: 'default') do
```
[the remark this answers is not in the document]
dmccown: `'name' => 'default'` is used in `.where` as the table data is keyed based on strings. In this case we are calling a constructor that can use a hash with symbols.

docs/resources/azure_security_center_policy.md.erb
```diff
@@ -0,0 +1,206 @@
...
+### Test the default Security Center Policy exists
+
+    describe azure_security_center_policy(name: 'default') do
+      it { should exist }
+    end
+
+### Test the default Security Center Policy has log collection enabled
+
+    describe azure_security_center_policy(name: 'default') do
+      its('log_collection') { should eq('On') }
```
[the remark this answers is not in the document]
dmccown: Yup, in this case `On` is capitalized.

docs/resources/azure_security_center_policy.md.erb
```diff
@@ -0,0 +1,206 @@
...
+
+    its('id') { should eq('/subscriptions/<SUBSCRIPTION_ID>/providers/Microsoft.Security/policies/default') }
+
+### name
+
+The name of the Security Center Policy.
+
+    its('name') { should eq('default') }
+
+### log_collection
```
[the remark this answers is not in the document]
dmccown: I think we only need to escape when there's two underscores. E.g. `my_log_collection` would add emphasis to that string. A single underscore shouldn't change the style as far as I know. Is there another reason we need to escape the underscore?

docs/resources/azure_security_center_policy.md.erb
```diff
@@ -0,0 +1,206 @@
...
+
+Log collection indicates if the monitoring agent will collect security data (`On`|`Off`).
+
+    its('log_collection') { should eq('On') }
```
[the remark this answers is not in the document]
dmccown: We could add a new matcher if that would be more consistent. I'd recommend we do that as a followup.

docs/resources/azure_security_center_policies.md.erb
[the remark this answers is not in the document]
jerryaldrichiii: Not sure, I just notice we mentioned a filter then don't give an example. Up to you :smile:

docs/resources/azure_security_center_policy.md.erb
```diff
@@ -0,0 +1,206 @@
...
+### contact_phone
+
+Contact phone contains the security contact phone number.
+
+    its('contact_phone') { should eq('1-111-111-1111') }
+
+## Matchers
+
+This InSpec audit resource has the following special matchers. For a full list
+of available matchers, please visit our [Universal Matchers
```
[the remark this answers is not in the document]
dmccown: Fixed to be safe.

docs/resources/azure_security_center_policies.md.erb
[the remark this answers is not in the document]
jerryaldrichiii: Yeah, I use `Assert that` but `Insist that` would be fine here I think.

docs/resources/azure_security_center_policy.md.erb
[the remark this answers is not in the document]
jerryaldrichiii: Ah, sounds good.

docs/resources/azure_security_center_policy.md.erb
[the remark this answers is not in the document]
jerryaldrichiii: I've done it with the other docs even if there is just one underscore. I'm not sure to be honest, but we escape all underscores in the core docs.

docs/resources/azure_security_center_policy.md.erb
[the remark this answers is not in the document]
jerryaldrichiii: Yup, same. Won't block this PR over it. :smile:

docs/resources/azurerm_virtual_machine.md.erb
```diff
@@ -0,0 +1,173 @@
...
+    end
+
+<br />
+
+## Examples
+
+### Ensure that the virtual machine has the expected data disks
+
+    describe azurerm_virtual_machine(resource_group: 'MyResourceGroup', name: 'MyVmName') do
+      its('data_disks.sort') { should eq ['DataDisk1', 'DataDisk2'] }
```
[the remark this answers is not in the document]
dmccown: You could, but that doesn't assert the same thing. The example I have would assert that you only have those two entries returned. 

You example would still be passing if there are other disks returned. We'd need to add another assertion on the number of disks returned.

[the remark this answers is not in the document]
jerryaldrichiii: This would avoid having to use `.sort`

docs/resources/azure_security_center_policy.md.erb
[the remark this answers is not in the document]
dmccown: Discussed and decided it's best to escape so we can be consistent.

docs/resources/azurerm_virtual_machine.md.erb
```diff
@@ -0,0 +1,173 @@
...
+    end
+
+### Ensure that the virtual machine has the expected monitoring agent installed
+
+    describe azurerm_virtual_machine(resource_group: 'MyResourceGroup', name: 'MyVmName') do
+      its('monitoring_agent_installed') { should eq true }
```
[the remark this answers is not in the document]
dmccown: I actually already have a matcher, so I can do:

```
it { should have_monitoring_agent_installed }
```
I'll switch the example.

docs/resources/azurerm_virtual_machine_disk.md.erb
```diff
@@ -0,0 +1,129 @@
...
+
+    /subscriptions/{subscriptionId}/resourceGroups/myResourceGroup/providers/Microsoft.Compute/disks/myManagedDisk
+
+### name
+
+The disk's name.
+
+    its('name') { should eq('MyDiskName') }
+
+### managed by
```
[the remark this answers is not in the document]
dmccown: I think this should be `managedBy` to be accurate with the implementation. The implementation should change so it's `managed_by` though.

libraries/azure_monitor_log_profile.rb
dmccown: `!!` converts a value into a boolean. So a falsy value such as `nil` would be false. A truthy value will come back as true.

libraries/azure_network_security_group.rb
dmccown: So something like?
```
ports.select { .... }
         .collect { .... }
         .any? { ... }
```

libraries/azure_security_center_policy.rb
dmccown: Generally we've used the HEREDOC syntax (`<<-EXAMPLE`). I think we should standardize on that instead of the multi-line string.

libraries/azurerm_virtual_machine_disk.rb
dmccown: See explanation: https://github.com/inspec/inspec-azure/pull/63#discussion_r199280731

test/functional/verify/controls/azure_network_security_group.rb
dmccown: I think so. Is there a spot where it's missing?

libraries/azurerm_virtual_machine.rb
dmccown: I'll remove. We resolved it by switching to `azurerm` instead of `azure`. We should probably have a comment on why the name is different or standardize on `azurerm`.

libraries/azurerm_virtual_machine_disk.rb
dmccown: https://github.com/inspec/inspec-azure/pull/63#discussion_r199281985

test/functional/verify/controls/azure_network_security_group.rb
dmccown: Ahh I see. Fixed.

libraries/azure_security_center_policy.rb
dmccown: I fixed so we use HEREDOCs everywhere.

docs/resources/azurerm_virtual_machine.md.erb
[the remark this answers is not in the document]
dmccown: I'll switch it.

docs/resources/azurerm_virtual_machines.md.erb
dmccown: This one is tricky as we have an array of arrays here. The underlying structure is:
```
[['MyDisk1', 'MyDisk2'], ['MyDisk3']]
```
Do you have a recommendation on how to do this more cleaning in the filter table implementation?

libraries/support/azure/management.rb
dmccown: In this case it's trying to say you may raise an error if things aren't configured correctly.

docs/resources/azurerm_virtual_machine_disk.md.erb
```diff
@@ -0,0 +1,129 @@
...
+
+### sku
+
+The disk's sku name (`Standard_LRS`|`Premium_LRS`).
+
+    its('sku') { should eq('Standard_LRS') }
+
+### properties
+
+The disk's properties. This is a hash of key/value pairs.
```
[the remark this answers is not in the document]
dmccown: I'll need to noodle on this. We can follow up post merge with an example if that's acceptable.

docs/resources/azure_security_center_policies.md.erb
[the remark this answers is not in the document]
dmccown: I've generally followed the pattern of the syntax block just shows the describe block without any content. I felt having a syntax example and then an example block got redundant.

libraries/support/azure/management.rb
jerryaldrichiii: Hmm. Fair enough. My rubyisms aren't the best.

docs/resources/azure_security_center_policies.md.erb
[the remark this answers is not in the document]
jerryaldrichiii: Sounds good.

docs/resources/azurerm_virtual_machine.md.erb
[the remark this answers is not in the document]
jerryaldrichiii: Ah, good point. My second comment came before I read yours.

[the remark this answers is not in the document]
jerryaldrichiii: I'd like to avoid sort though. A count would work, but I'll leave it to you.

docs/resources/azurerm_virtual_machines.md.erb
jerryaldrichiii: Hmmm...not at the moment, you're right. Array of array's is tricky. Disregard my comment :smile:.

libraries/azure_network_security_group.rb
jerryaldrichiii: Yip :smile:

libraries/azure_security_center_policy.rb
jerryaldrichiii: You da best.

inspec.lock
```diff
@@ -0,0 +1,3 @@
+---
```
clintoncwolfe: We should not ship the inspec lockfile with the resource pack.

inspec.yml
```diff
@@ -1,7 +1,11 @@
...
-copyright: russell.seymour@turtlesystems.co.uk
-copyright_email: russell.seymour@turtlesystems.co.uk
-version: 0.2.0
\ No newline at end of file
+maintainer: Chef Software, Inc.
+copyright: Chef Software, Inc.
+copyright_email: support@chef.io
+license: Apache-2.0
+summary: This resource pack provides compliance resources for Azure.
+version: 1.0.0
```
clintoncwolfe: :100: for the semver major bump :)

test/functional/verify/controls/azure_network_security_group.rb
```diff
@@ -0,0 +1,24 @@
+resource_group = attribute('resource_group',            default: nil)
```
clintoncwolfe: For consistency, we may eventually want to reclass these tests as "integration" tests (that's what the inspec project calls any test that does live-fire tests against a specific platform target).  inspec project uses the word 'functional' to refer specifically to UX testing.  Not a blocker issue.

bin/inspec
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env ruby
```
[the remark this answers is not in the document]
dmccown: This was needed when we had continuous integration setup. If you have `chefdk` installed you can end up in a situation where the version of InSpec you bundle with will be different that the version of InSpec included in ChefDK. The binstub guarantees that you will always run with the version of InSpec you are bundled with.

docs/resources/azurerm_virtual_machine.md.erb
```diff
@@ -0,0 +1,172 @@
+---
```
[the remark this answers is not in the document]
dmccown: At the moment it was just to avoid an implementation collision between the existing azure_virtual_machine resource and this implementation. I can update the docs to make that more clear, or we could rename all the new resources to the `azurerm` namespace if that would be more clear.

`azurerm` seems the preferred namespace in Terraform.

docs/resources/azure_generic_resource.md
```diff
@@ -1,81 +0,0 @@
----
```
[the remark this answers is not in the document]
dmccown: What's the risk if we have an implementation in `inspec-core` and the implementation in the resource pack? If they diverge which implementation will the resource pack use?

test/functional/verify/controls/azure_network_security_group.rb
dmccown: We can rename the folder that's not a big deal.

docs/resources/azurerm_virtual_machine.md.erb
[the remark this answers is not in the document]
trevorghess: azure vs azurerm is a common thing. It denotes the separation between Azure classic resource (Azure Service Manager or ASM) vs Azure Resource Manager (ARM)

jerryaldrichiii:
All of my feedback is addressed thanks @dmccown!

PR #70, opened 2018-07-04
Adds azurerm_ad_user and azurerm_ad_users resources

r-fennell:
Add Resources, enpoints and control(s) to satisfy CIS 1.3 benchmark ensuring no AD Guest user accounts in azure.


libraries/azure_ad_users.rb
```diff
@@ -0,0 +1,56 @@
...
+                          displayName:  user["displayName"],
+                          mail:         user["mail"],
+                          userType:     user["userType"]
+                      }]
+      end
+      next_page = @users["nextLink"]
+      break unless next_page
+    end
+    @table = user_rows
+  end
```
TrevorBramble: Let's revisit this code after the pagination refactoring discussed elsewhere. Long & short of it is that the above should be something more like:
```ruby
def initialize
  @table = graph_client.users
end
```
_Maybe_
```ruby
USER_KEYS = %i(objectId displayName mail userType).freeze

# ... FilterTable and whatnot

def initialize
  @table = graph_client.users.map(&user_to_hash)
end

def user_to_hash
  ->(u) { USER_KEYS.zip(u.values_at(USER_KEYS.map(&:to_s)).to_hash) }
end
```
...but I believe this is actually contrary to the established pattern, where we've kept parity with the AWS implementation and collections of resources are simply lists of identifiers.

libraries/azure_ad_users.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def to_s
+    "Azure AD Users"
```
TrevorBramble: Could bear spelling out the whole thing. =^)

libraries/azure_ad_user.rb
```diff
@@ -0,0 +1,47 @@
...
+    mail
+    displayName
+    userType
+  ).freeze
+
+  attr_reader(*ATTRS)
+
+  # Can be constructed either via API call with an Azure ID provided,
+  # or from parameters passed. API call takes precedence when ID provided.
+  def initialize(azure_id, user_args)
```
TrevorBramble: In the convention we've been using, this should only ever take an ID. (See comments elsewhere about collections just containing IDs.)

libraries/azure_ad_users.rb
r-fennell: Sure - I totally agree the pagination should be dealt with at a level below this, I just wasn't sure where best suited. Can you link me to the previous discussions mentioned so I can get up to speed? Will leave how it is until we have chatted and I have a clearer picture of where the logic belongs.

libraries/support/azure/graph.rb
```diff
@@ -0,0 +1,66 @@
...
+
+require 'json'
+require 'singleton'
+
+module Azure
+  class Graph
+    include Singleton
+
+    def with_client(azure_client, override: false)
+      set_reader(:azure_client, azure_client, override)
```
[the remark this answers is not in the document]
r-fennell: :sweat_smile:

libraries/support/azure/graph.rb
```diff
@@ -0,0 +1,66 @@
...
+    private
+
+    def get(*args)
+      confirm_configured!
+      response          = azure_client.get(*args).body
+      map               = {}
+      map["values"]     = response.fetch('value', response)
+      map["nextLink"] ||= response.fetch("odata.nextLink", nil)
+      map
+    end
```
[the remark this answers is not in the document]
r-fennell: Agree, this is perhaps leaving too much to dev knowledge. It is linked to the "where should we paginate" you mentioned above and so similarly I will wait until we have discussed further.

>the get definition should also line up with what's already in Management. (Keyword arguments :url, :api_version.)
I have seen in some of the MS documentation that there are places where params other than `api_version` would be very beneficial (e.g. providing a filter to only get certain entities, instead of getting all entities and hand rolling the logic), is there a reason not to allow a caller to provide their own hash of params instead of limiting to only `api-version`?

libraries/support/azure/rest.rb
```diff
@@ -5,11 +5,10 @@
 
 module Azure
   class Rest
-    DEFAULT_HOST = 'https://management.azure.com'
 
     attr_reader :host, :resource, :credentials
 
-    def initialize(host: DEFAULT_HOST, credentials: {})
+    def initialize(host, credentials: {})
```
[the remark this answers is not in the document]
r-fennell: My thinking here was to abstract the logic of which API is being used behind the `azurerm_resource.rb` convenience methods (e.g. `client`/`graph_client`) and it didnt make sense to prefer one API over another.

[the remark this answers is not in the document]
TrevorBramble: I agree. But that doesn't mean we have to drop the keyword arg, only its default.  `def initialize(host:, credentials: {})`

TrevorBramble:
@r-fennell I still see some things that won't pass our Rubocop criteria. Please be sure to run `rake lint`.

libraries/azure_ad_users.rb
```diff
@@ -0,0 +1,32 @@
...
+require 'azurerm_resource'
+
+class AzureAdUsers < AzurermResource
+  name 'azure_ad_users'
+  desc 'Verifies settings for a collection of Azure Active Directory Users'
+  example "
+    describe azure_ad_users do
+        it  { should exist }
+    end
+  "
```
TrevorBramble: Heredoc for all example strings

docs/resources/azurerm_ad_users.md
```diff
@@ -0,0 +1,100 @@
+---
+title: About the azure_resource_groups Resource
```
pwelch: I think this should be "About the `azure_ad_users` Resource"

docs/resources/azurerm_ad_users.md
```diff
@@ -0,0 +1,100 @@
...
+* `names`
+
+### names
+
+Filters the results to include only those resource groups that match the given
+name. This is a string value.
+
+     describe azurerm_ad_users.where{ displayName.eql?('Joe Bloggs') } do
+      it { should exist }
+    end
```
pwelch: Nitpick, looks like `describe` is one space over too far

```
    describe azurerm_ad_users.where{ displayName.eql?('Joe Bloggs') } do
      it { should exist }
    end
```

docs/resources/azurerm_ad_users.md
```diff
@@ -0,0 +1,100 @@
...
+
+### azureIds
+
+The azureIds property provides a list of all User's Azure IDs.
+
+    its('azure_ids') { should include '44211066-f292-4546-8ced-2ab0e0911f44' }
+    
+### displayNames
+
+The displayNames property provides a list of all the User display names.
```
pwelch: Should we match this with to say `display_names` like the example or should it be `displayNames`?

docs/resources/azurerm_ad_users.md
```diff
@@ -0,0 +1,100 @@
...
+
+    its('display_names') { should include 'Azure Admin Account' }
+    
+### mails
+
+The mail property provides a list of all the User email addresses, where present.
```
pwelch: Change `mail` to `mails` ?

docs/resources/azurerm_ad_users.md
r-fennell: The column in the FilterTable is indeed `display_names` - will update example!

libraries/azurerm_ad_users.rb
```diff
@@ -0,0 +1,59 @@
...
+        it  { should exist }
+    end
+  EXAMPLE
+
+  filter = FilterTable.create
+  filter.register_column(:azure_ids,      field: 'azureId')
+  filter.register_column(:display_names,  field: 'displayName')
+  filter.register_column(:mails,          field: 'mail')
+  filter.register_column(:user_types,     field: 'userType')
+  filter.install_filter_methods_on_resource(self, :data)
```
dmccown: Nit: you can change `:data` to `:table`.

libraries/azurerm_ad_users.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def initialize
+    @table = graph_client.users
+
+    @exists = !@table.empty?
+  end
+
+  def data
```
dmccown: This method may go away.

libraries/azurerm_ad_users.rb
```diff
@@ -0,0 +1,59 @@
...
+    @table
+  end
+
+  def guest_accounts
+    GuestUsers.new(@table)
```
dmccown: I recommend memoization here. 
```
@guest_users ||= GuestUsers.new(@table)
```

libraries/azurerm_ad_users.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def to_s
+    'Azure Active Directory Users'
+  end
+
+  class GuestUsers
+    def initialize(all_users)
+      @guests = []
+
+      all_users.each do |user|
```
dmccown: You may also write this with a map and get the table for free.
```
@guests = all_users.select { |u| user['userType'] == 'Guest' }
                   .map { |u| { displayName: user['displayName'], mail: user['mail'], userType: user['userType'] } }
```

libraries/azurerm_resource.rb
```diff
@@ -5,26 +5,39 @@
...
+  GRAPH_HOST = 'https://graph.windows.net'
+
   def exists?
     @exists ||= false
   end
 
   def client
     Azure::Management.instance
-                     .with_client(azure_client)
+                     .with_client(rest_client(MGMT_HOST))
```
dmccown: I think leaving a default for `MGMT_HOST` is a reasonable default. I'd also recommend spelling out `MANAGEMENT_HOST`.

libraries/support/azure/graph.rb
```diff
@@ -0,0 +1,83 @@
...
+    def users
+      get(
+        "/#{tenant_id}/users",
+        params: { 'api-version' => '1.6' },
+      )
+    end
+
+    private
+
+    def get(*args)
```
dmccown: This method still feels like it's doing a lot of work. Can we split out the pagination logic to it's own class?

libraries/support/azure/rest.rb
```diff
@@ -5,11 +5,9 @@
 
 module Azure
   class Rest
-    DEFAULT_HOST = 'https://management.azure.com'
-
     attr_reader :host, :resource, :credentials
 
-    def initialize(host: DEFAULT_HOST, credentials: {})
+    def initialize(host, credentials: {})
```
dmccown: Defaulting to management seems reasonable to me as most of our resources will speak to management by default.

test/integration/verify/controls/azure_network_watcher.rb
```diff
@@ -1,6 +1,6 @@
 resource_group = attribute('resource_group',       default: nil)
-nw             = attribute('network_watcher_name', default: nil).first
-nw_id          = attribute('network_watcher_id',   default: nil).first
+nw             = attribute('network_watcher_name', default: nil) # .first TODO uncomment (breaking)
```
dmccown: We spoke about this being broken. If you tear down your environment and rebuild I think you'll find that you'll need to add the `.first` back to each of these lines.

test/integration/verify/controls/azurerm_ad_users.rb
```diff
@@ -0,0 +1,9 @@
+control 'azure_ad_users' do
+
+  title 'Ensure no Guest Users are present within an Active Directory Tenant'
+
+  describe azurerm_ad_users.guest_accounts do
+    it { should_not exist }
```
dmccown: Is guest_accounts `nil` or `[]` (empty list)?

.rubocop.yml
```diff
@@ -69,7 +69,9 @@ Style/GuardClause:
   Enabled: false
 Style/IfUnlessModifier:
   Enabled: false
-Style/MethodMissing:
+Style/MethodMissingSuper:
+  Enabled: false
```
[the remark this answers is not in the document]
r-fennell: When running rubocop it would complain about this check:

```rake lint
Running RuboCop...
Error: The `Style/MethodMissing` cop has been split into `Style/MethodMissingSuper` and `Style/MissingRespondToMissing`.
(obsolete configuration found in .rubocop.yml, please update it)
RuboCop failed!```

I was getting the same issue on master

File: test/integration/verify/controls/azurerm_ad_users.rb (hunk 20 above)
Comment (r-fennell) [reply test/integration/verify/controls/azurerm_ad_users.rb]: It will return an object of type GuestAccount, with an `exists` method. I figured this would read more in the style of inspec.

Comment (dmccown) [reply test/integration/verify/controls/azurerm_ad_users.rb]: I'm not sure what the output would look like if it does exist. Would you get a list of users that are guest accounts? I'm just wondering how you'd know which accounts are guest accounts so you could fix the finding.

File: .rubocop.yml (hunk 21 above)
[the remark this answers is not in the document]
Comment (pwelch) [reply .rubocop.yml]: Possibly from a new version of Rubocop? Whenever I upgrade I always find new default lint errors.

File: test/integration/verify/controls/azurerm_ad_users.rb (hunk 20 above)
Comment (r-fennell) [reply test/integration/verify/controls/azurerm_ad_users.rb]: Output on fail would look like:
```
×  azure_ad_users: Ensure no Guest Users are present within an Active Directory Tenant
     ×  [
       {
         "displayName": "User One",
         "mail": "one@email.com",
         "userType": "Guest"
       },
       {
         "displayName": "User Two",
         "mail": "two@email.com",
         "userType": "Guest"
       }
     ] should not exist
     expected #<#<Class:0x00007f8eb52a8028>::AzurermAdUsers::GuestUsers:0x00007f8eb99fb6f0 @guests=[{ ... }], @exists=true> not to exist
```
Giving you enough information for remediation.
If no Guest users are active, exists will be false and the control will pass.

File: libraries/azurerm_ad_users.rb [hunk 22]
```diff
@@ -0,0 +1,56 @@
...
+
+  class GuestUsers
+    def initialize(all_users)
+      @guests = all_users.select { |user| user['userType'] == 'Guest' }
+                         .map { |user|
+                           {
+                             displayName: user['displayName'],
+                               mail:        user['mail'],
+                               userType:    user['userType'],
+                           }
```
pwelch: Superficial change but could you align this on the `:` unless Rubocop complains about it.

```
{
{
  displayName: user['displayName'],
         mail: user['mail'],
     userType: user['userType'],
}
```

docs/resources/azurerm_ad_users.md.erb
```diff
@@ -0,0 +1,106 @@
...
+### user\_types
+
+The userTypes property provides a list of all User Types for all users.
+
+    its('user_types') { should include 'Member' }
+
+## Matchers
+
+This InSpec audit resource has the following special matchers. For a full list
+of available matchers, please visit our [Universal Matchers
```
dmccown: Awkward wrap here. Will that break Markdown?

libraries/azurerm_ad_users.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'azurerm_resource'
+require 'json'
+
+class AzurermAdUsers < AzurermResource
+  name 'azurerm_ad_users'
+  desc 'Verifies settings for a collection of Azure Active Directory Users'
+  example <<-EXAMPLE
+    describe azure_ad_users do
```
dmccown: Resource name is wrong. It should be `azurerm_ad_users`.

libraries/azurerm_ad_user.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require 'azurerm_resource'
+
+class AzurermAdUser < AzurermResource
+  name 'azurerm_ad_user'
+  desc 'Verifies settings for an Azure Active Directory User'
+  example <<-EXAMPLE
+    describe azure_ad_user(azureId: 'azure_id') do
```
dmccown: Resource name is wrong. It should be `azurerm_ad_user`.

dmccown: This example is wrong too. Is this the right way to call this resource? If we use a hash I think the constructor will need to change.

docs/resources/azurerm_ad_users.md.erb
```diff
@@ -0,0 +1,106 @@
+---
+title: About the azurerm_ad_users Resource
+platform: azure
+---
+
+# azurerm\_ad\_users
+
+Use the `azurerm_ad_users` InSpec audit resource to test properties of
+some or all Azure Active Directory users within a Tenant.
+<br />
```
Padgett: Pattern is to have an empty line above and below

docs/resources/azurerm_ad_users.md.erb
```diff
@@ -0,0 +1,106 @@
...
+
+## Syntax
+
+An `azurerm_ad_users` resource block returns all Azure Active Directory user accounts for the
+contained within the configured Tenant and then tests that group group of users.
+
+    describe azurerm_ad_users do
+      ...
+    end
+<br />
```
Padgett: Here too

docs/resources/azurerm_ad_users.md.erb
```diff
@@ -0,0 +1,106 @@
...
+    describe azurerm_ad_users do
+      it { should exist }
+    end
+
+### Ensure there are no Guest accounts active
+
+    describe azurerm_ad_users.guest_accounts
+      it { should_not exist }
+    end
+<br />
```
Padgett: Need a blank line above

dmccown:
We should also add an integration test for the resource `azurerm_ad_user`.

libraries/azurerm_ad_user.rb
Padgett: This doesn't match the `initialize` method arguments. Suggest either removing the declaration and just passing in the id or changing the method to match.

Also, would `user_id` be a better choice?

libraries/azurerm_ad_user.rb
```diff
@@ -0,0 +1,41 @@
...
+  ATTRS = %i(
+    azureId
+    mail
+    displayName
+    userType
+  ).freeze
+
+  attr_reader(*ATTRS)
+
+  def initialize(id)
```
Padgett: This does not match the `example` given. See line 9 for additional comment

libraries/azurerm_ad_user.rb
dmccown: In the response on the plural resource the field is called `ObjectId`.

libraries/azurerm_ad_user.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require 'azurerm_resource'
+
+class AzurermAdUser < AzurermResource
+  name 'azurerm_ad_user'
+  desc 'Verifies settings for an Azure Active Directory User'
+  example <<-EXAMPLE
+    describe azure_ad_user(user_id: 'userId') do
```
TrevorBramble: Should be `azurerm_ad_user`

libraries/azurerm_resource.rb
```diff
@@ -5,26 +5,39 @@
...
+    Azure::Graph.instance
+                .with_client(rest_client(GRAPH_HOST))
+                .for_tenant(tenant_id)
+  end
+
   private
 
-  def azure_client
-    Azure::Rest.new(credentials: credentials.to_h)
+  def rest_client(host)
```
TrevorBramble: Default to management?

libraries/support/azure/rest.rb
```diff
@@ -5,11 +5,9 @@
 
 module Azure
   class Rest
-    DEFAULT_HOST = 'https://management.azure.com'
-
     attr_reader :host, :resource, :credentials
 
-    def initialize(host: DEFAULT_HOST, credentials: {})
+    def initialize(host = MANAGEMENT_HOST, credentials: {})
```
TrevorBramble: How is `MANAGEMENT_HOST` in scope?

PR #73, opened 2018-07-06
Add Availability section in each doc

### LINKED ISSUE #64: Add Availability section to each resource docs
A section should be added indicating that the resource is available via a resource pack, how to install, and what version the resource was first available.

clintoncwolfe:
In support of displaying the Azure resource pack docs on the main InSpec website, we need to note that these resources are distributed as a resource pack.  This PR adds such a note.

Additionally, we get a lot of confusion around when a resource first appeared (and thus minimum version needed).  This PR also adds a note to that affect, as well.

screen shot 2018-07-05 at 22 36 18

Fixes #64 #93

### DIFF
--- a/docs/resources/azurerm_ad_user.md.erb
+++ b/docs/resources/azurerm_ad_user.md.erb
@@ -1,0 +1,184 @@
+---
+title: About the azurerm_ad_user Resource
+platform: azure
+---
+
+# azurerm\_ad\_user
+
+Use the `azurerm_ad_user` InSpec audit resource to test properties of
+an Azure Active Directory user within a Tenant.
+
+<br />
+
+## Availability
+
+### Installation
+
+This resource is available in the `inspec-azure` [resource
+pack](https://www.inspec.io/docs/reference/glossary/#resource-pack). To use it, add the
+following to your `inspec.yml` in your top-level profile:
+
+    depends:
+      inspec-azure:
+        git: https://github.com/inspec/inspec-azure.git
+
+You'll also need to setup your Azure credentials; see the resource pack
+[README](https://github.com/inspec/inspec-azure#inspec-for-azure).
+
+### Version
+
+This resource first became available in 1.1.0 of the inspec-azure resource pack.
+
+## Syntax
+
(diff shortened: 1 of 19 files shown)


docs/resources/azure_resource_groups.md.erb
```diff
@@ -13,6 +13,20 @@ common command on a group of resources.
 
 <br />
 
+## Availability
+
+This resource is available in the `inspec-azure` [resource pack](https://www.inspec.io/docs/reference/glossary/#resource-pack).  To use it, add the following to your `inspec.yml` in your top-level profile:
+
+    depends:
+      inspec-azure:
+        git: https://github.com/inspec/inspec-azure.git
```
dmccown: Does it make any different if we point people to the repo or the archive file off of that repo? E.g. `git: https://github.com/inspec/inspec-azure/archive/master.tar.gz`

clintoncwolfe: I'm not sure.  We may be able to use the http/tarball approach and relax the need to have git installed.

clintoncwolfe:
FYI the Availability section is mechanically generated and may be updated dynamically as well.

docs/resources/azure_resource_groups.md.erb
dmccown: I tested with the tarball and it was working with that instead of the git URI.

TrevorBramble:
@clintoncwolfe It is very inconvenient to review unwrapped lines. What makes it hard to wrap the lines?

clintoncwolfe:
Welp, I thought the width was hardcoded in commonmark, but I just found where to change it in Ruby.  I'll special-case inspec-azure to 90.

clintoncwolfe:
Additional screencaps (made by running the website release procedure on the cw/contrib-pull-docs branch of the inspec/inspec repo, with the contrib.yaml referencing the cw/doc-add-availability branch of the inspec-azure repo). 

screen shot 2018-07-12 at 15 19 54
screen shot 2018-07-12 at 15 20 08
screen shot 2018-07-12 at 15 20 46

PR #75, opened 2018-07-09
Add azure_network_watcher(s) Resource Documentation

Padgett:
- Add azure_network_watchers resource documentation
- Add azure_network_watcher resource documentation

### DIFF
--- a/docs/resources/azure_network_watcher.md.erb
+++ b/docs/resources/azure_network_watcher.md.erb
@@ -1,0 +1,84 @@
+---
+title: About the azure_network_watcher Resource
+platform: azure
+---
+
+# azure\_network\_watcher
+
+Use the `azure_network_watcher` InSpec audit resource to test properties
+of an Azure Network Watcher.
+
+<br />
+
+## Syntax
+
+An `azure_network_watcher` resource block identifies a Network Watcher by
+name and resource group.
+
+    describe azure_network_watcher(resource_group: 'example', name: 'WatcherName') do
+      ...
+    end
+<br />
+
+## Examples
+
+### Test that an example Resource Group has the specified Network Watcher
+
+    describe azure_network_watcher(resource_group: 'example', name: 'WatcherName') do
+      it { should exist }
+    end
+
+### Test that an example Resource Group has a Network Watcher with the Provisioning State value 'Succeeded'
+
+    describe azure_network_watcher(resource_group: 'example', name: 'WatcherName') do
+      its('provisioning_state') { should eq 'Succeeded' }
+    end
+
+<br />
+
+## Parameters
+
+* `name`
+* `resource_group`
+
+## Parameter Examples
+
+The Resource Group as well as the Network Watcher name.
+
+    describe azure_network_watcher(resource_group: 'example', name: 'WatcherName') do
+      its('provisioning_state') { should eq 'Succeeded' }
+    end
+
+## Properties
+
+* `provisioning_state`
+
+### provisioning_state
+
+The provisioning_state field can be checked for the value of the Provisioning State.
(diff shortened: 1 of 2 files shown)


docs/resources/azure_network_watcher.md.erb
```diff
@@ -0,0 +1,84 @@
...
+
+## Examples
+
+### Test that an example Resource Group has the specified Network Watcher
+
+    describe azure_network_watcher(resource_group: 'example', name: 'WatcherName') do
+      it { should exist }
+    end
+
+### Test that an example Resource Group has a Network Watcher with the correct Provisioning State
```
TrevorBramble: The examples should be a little less dry. In this one, we're specifically testing that the watcher was successfully provisioned, so let's say that instead of the more generic "correct provisioning state".

docs/resources/azure_network_watcher.md.erb
```diff
@@ -0,0 +1,84 @@
...
+
+    describe azure_network_watcher(resource_group: 'example', name: 'WatcherName') do
+      its('provisioning_state') { should eq 'Succeeded' }
+    end
+
+## Properties
+
+* `provisioning_state`
+
+### provisioning_state
```
dmccown: For consistency let's escape the `_` (`\_`).

docs/resources/azure_network_watcher.md.erb
```diff
@@ -0,0 +1,84 @@
...
+
+The provisioning_state field can be checked for the value of the Provisioning State.
```
dmccown: I'd suggest escaping the underscore here as well. If another underscore appears on this line formatting will get messed up.

docs/resources/azure_network_watcher.md.erb
```diff
@@ -0,0 +1,84 @@
...
+
+    # If we expect 'WatcherName' to always exist
+    describe azure_network_watcher(resource_group: 'example', name: 'WatcherName') do
+      it { should exist }
+    end
+
+    # If we expect 'WatcherName' to never exist
+    describe azure_network_watcher(resource_group: 'example', name: 'WatcherName') do
+      it { should_not exist }
+    end
```
dmccown: I'd suggest using another name for the negative example. Maybe something like `WatcherNotFound`. Just to clear up that these are different use cases.

docs/resources/azure_network_watchers.md.erb
```diff
@@ -0,0 +1,63 @@
...
+
+This InSpec audit resource has the following special matchers. For a full list
+of available matchers, please visit our [Universal Matchers page](https://www.inspec.io/docs/reference/matchers/).
+
+### exists
+
+The control will pass if the resource returns a result. Use `should_not` if you expect zero matches.
+
+    # If we expect 'example_group' Resource Group to have Network Watchers
+    describe azure_network_watchers(resource_group: 'example_group') do
```
dmccown: We should use mixed case instead of snake case. `ExampleGroup` would be my recommendation.

docs/resources/azure_network_watchers.md.erb
```diff
@@ -0,0 +1,63 @@
...
+      it { should exist }
+    end
+
+    # If we expect 'example_group' Resource Group to not have Network Watchers
+    describe azure_network_watchers(resource_group: 'example_group') do
```
dmccown: We should use mixed case instead of snake case. `ExampleGroup` would be my recommendation.

dmccown: I'd also suggest a different name here so it's clear this group doesn't exist. E.g. `MissingExampleGroup`.

PR #77, opened 2018-07-09
Replaces connection with InSpec backend

dmccown:
This change replaces our connection logic with the InSpec backend
(backed by Train) to handle connections. By doing so, we can now support
using a credentials file, MSI connector as well as shell vars.

I'm also adding another terraform target to start up a linux virtual
machine with MSI enabled. You will need to add that MSI account with the
contributor role to your subscription for MSI to work against the REST
API.

To start up the MSI vm:

`rake msi_vm tf:apply`

You will need to add a public ssh key to your `.envrc` file as well. An
example is included in `.envrc-example`.

Note: If you run integration tests while you have the MSI vm running
tests will fail. The tests were written with the assumption that there
would be only two vms.


dmccown:
I can look at making the integration tests pass with the extra vm running...

dmccown:
Now that #70 has been merged we will need to introduce a change that will allow us to talk to different Azure resource endpoints. 

As this branch is right now it cannot Authenticate with the Graph API. 

https://github.com/inspec/inspec-azure/blob/dmccown/replace-connection-with-train/libraries/azurerm_resource.rb#L34

Is an instance of a [MsRest::TokenCredentials](https://github.com/Azure/azure-sdk-for-ruby/blob/master/runtime/ms_rest/lib/ms_rest/credentials/token_credentials.rb) with settings for how to authenticate against Management API. 

We need a way to get a different instance of `MsRest::TokenCredentials` from Train with the correct settings found here: https://github.com/Azure/azure-sdk-for-ruby/blob/master/data/azure_graph_rbac/spec/1.6/spec_helper.rb#L22-L27. In lines 23 and 24 the `authentication_endpoint` and `token_audience` are changed so it can authenticate against the Graph API.

We'll need to look at changing [Train::Transports::Azure](https://github.com/inspec/train/blob/master/lib/train/transports/azure.rb) to provide that new credentials object.

Furthermore if a user would like to use the Ruby SDK implementation to communicate with the Graph API they would need an instance of `GraphRbacClient`. An example usage can also be found here: https://github.com/Azure/azure-sdk-for-ruby/blob/master/data/azure_graph_rbac/spec/1.6/spec_helper.rb#L36-L38. *Note:* `tenant_id` must be manually set after the object is instantiated. 

## Other thoughts

We'll need to be sure any changes we want to make within Train will not break existing users. We should not change the public interface in a breaking manner.

I'd like to see as much of the connection logic in Train as possible.

## Testing locally
In Train add the follow to `train.gemspec`
```
spec.add_dependency 'azure_graph_rbac'
```
```
$ bundle
```
In your resource pack add this to your `Gemfile`
```
gem 'train', path: 'your/path/to/local/train/checkout'
```
```
$ bundle
```

You may now make changes to Train locally and see them reflected in the resource pack.

Gemfile
```diff
@@ -6,6 +6,7 @@ gem 'faraday',            '~> 0.15.0'
 gem 'faraday_middleware', '~> 0.12.2'
 gem 'inspec',             '~> 2.2',  '>= 2.2.7'
 gem 'rake',               '~> 12.3', '>= 12.3.1'
+gem 'train', path: '~/repos/train'
```
dmccown: Let's make sure this doesn't get included in the final PR.

README.md
```diff
@@ -289,6 +289,20 @@ direnv allow # or source .envrc
 rake tf:apply
 ```
 
+#### Managed Service Identity
+
+Managed Service Identity (MSI) is another way to connect to the Azure APIs.
+This option starts an additonal virtual machine with MSI enabled and a public
+ip address. You will need to put a hole in your firewall to connect to the
+virtual machine. You will also need to grant the `contributor` role to this
+identity for your subscription.
```
r-fennell: Should we include instructions on how to add this role? [Microsoft Docs](https://docs.microsoft.com/en-us/azure/role-based-access-control/role-assignments-portal)

dmccown: We describe that process earlier in the README.md https://github.com/inspec/inspec-azure/blob/9e1d5706950be5e05587d5d4829cb6dfd27b9992/README.md#service-principal

libraries/azurerm_ad_user.rb
```diff
@@ -52,6 +52,6 @@ def guest?
   end
 
   def to_s
-    "Azure Active Directory Username: '#{displayName}' with email '#{mail}'"
+    "Azure Active Directory Username: '#{displayName}' with objectId '#{objectId}'"
```
dmccown: Is the User Id a good thing to put here to identify the user? An email address would be easier to look up in your system wouldn't it?

libraries/azurerm_resource.rb
```diff
@@ -5,42 +5,36 @@
...
     Azure::Management.instance
-                     .with_client(rest_client)
+                     .with_client(management_client)
                      .for_subscription(subscription_id)
   end
 
-  def graph_client
+  def graph
+    raise Inspec::Exceptions::ResourceSkipped, 'MSI Authentication is currently unsupported '\
+      'for Graph RBAC API, control skipped.' if inspec.backend.msi_auth?
```
dmccown: Maybe:
```ruby
if inspec.backend.msi_auth?
  raise Inspec::Exceptions::ResourceSkipped, 'MSI Authentication is currently unsupported for Active Directory Users.'
end
```
Doing a guard clause puts the conditional at the front, which is the most important thing here. I reworded the error message so a user would have an easier time connecting that the resource isn't supported at the moment without needed to know what the Graph API is.

libraries/support/azure/rest.rb
```diff
@@ -7,18 +7,20 @@ module Azure
...
     attr_reader :host, :resource, :credentials
 
-    def initialize(host, credentials: {})
-      @host        = host
-      @resource    = "#{host}/"
-      @credentials = credentials
+    def initialize(client)
+      @host        = client.base_url
+      @resource    = client.base_url.to_s
+      @resource    += '/' unless @resource[-1, 1] == '/'
```
dmccown: What's the intent of this line?

libraries/azurerm_resource.rb
TrevorBramble: Putting this in a conditional structure would definitely make the condition prominent, but why not extract the conditional behavior to a method with an explaining name, to meet both interests?

```ruby
def graph
  skip_msi_unsupported if inspec.backend.msi_auth?

  # ...
end

def skip_msi_unsupported
  raise Inspec::Exceptions::ResourceSkipped,
        'MSI Authentication is currently unsupported for Active Directory Users.'
end
```

libraries/azurerm_ad_user.rb
r-fennell: Email would be nicer - but it is a nullable field and so leads to unreliable output

libraries/support/azure/rest.rb
r-fennell: Exactly - thanks for the cleaner version @TrevorBramble

PR #78, opened 2018-07-10
FilterTable syntax update

TrevorBramble:
Convert FilterTable to new syntax and make requisite dependency requirement change.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ source 'https://rubygems.org'
 
 gem 'faraday',            '~> 0.15.0'
 gem 'faraday_middleware', '~> 0.12.2'
-gem 'inspec'
+gem 'inspec',             '~> 2.2',  '>= 2.2.7'
 gem 'rake',               '~> 12.3', '>= 12.3.1'
 
 group :development do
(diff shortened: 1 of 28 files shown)


Gemfile
```diff
@@ -4,7 +4,7 @@ source 'https://rubygems.org'
 
 gem 'faraday',            '~> 0.15.0'
 gem 'faraday_middleware', '~> 0.12.2'
-gem 'inspec'
+gem 'inspec',             '~> 2.2',  '>= 2.2.7'
```
dmccown: 👍 Thanks!

libraries/azurerm_resource.rb
```diff
@@ -34,3 +30,11 @@ def credentials
     end
   end
 end
+
+class AzurermPluralResource < AzurermResource; end
+
+class AzurermSingularResource < AzurermResource
+  def exists?
```
dmccown: @clintoncwolfe can you recommend if we should do `exist` or `exists`?

TrevorBramble: @dmccown `exists?` is preferred, and I assume `exist?` is a fallback. https://github.com/inspec/inspec/blob/master/docs/dev/filtertable-usage.md#some-things-are-defined-for-you

clintoncwolfe: FilterTable, used for plural resources, uses the plural inflection, `exist?` 

Singular resources should use the singular inflection, `exists?`.

That's my take.

PR #85, opened 2018-07-11
Dmccown/readme cleanup

### LINKED ISSUE #72: README implies Terraform is required
As this project moves from a development project to a user-facing repository of resources, we need to make sure the README makes sense to multiple audiences.  One confusing point to me was the matter of Terraform and Direnv; if I were a user with an existing Azure infra I wanted to test, would I need to setup Terraform, manage workspaces, remote state, etc?

pwelch:
Fixes #72

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -189,12 +189,11 @@ end
 namespace :docs do
   desc 'Prints markdown links for resource doc files to update the README'
   task :resource_links do
-    files = Dir.entries('docs/resources').select { |f| !File.directory? f }
-
     puts "\n"
-    files.each do |file|
-      puts "- [#{file.split('.')[0]}](docs/resources/#{file})"
-    end
+    Dir.entries('docs/resources')
+       .select { |file| !File.directory?(file) }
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -185,3 +185,16 @@ namespace :tf do
...
+namespace :docs do
+  desc 'Prints markdown links for resource doc files to update the README'
+  task :resource_links do
+    files = Dir.entries('docs/resources').select { |f| !File.directory? f }
+
+    puts "\n"
+    files.each do |file|
+      puts "- [#{file.split('.')[0]}](docs/resources/#{file})"
+    end
+    puts "\n"
```
TrevorBramble: Might want to incorporate the whole thing in chained list processing instead of using local assignment for part of it.
```ruby
task :resource_links do
  puts "\n"
  Dir.entries('docs/resources')
     .select { |file| !File.directory?(file) }
     .collect { |file| "- [#{file.split('.')[0]}](docs/resources/#{file})" }
     .map { |link| puts link }
  puts "\n"
end
```

pwelch: Awesome idea, updated!

Issue #81, opened 2018-07-10
Resources need to be prefixed `azurerm`


Padgett:
Currently resources are mixed between `azure` and `azurerm` prefixes. Standardize on `azurerm` and alias `azure` with a deprecation warning.
Be sure to update integration tests and documentation.

- [x] azure_monitor_activity_log_alert
- [x] azure_monitor_activity_log_alerts
- [x] azure_monitor_log_profile
- [x] azure_monitor_log_profiles
- [x] azure_network_security_group
- [x] azure_network_security_groups
- [x] azure_network_watcher
- [x] azure_network_watchers
- [x] azure_resource_groups
- [x] azure_security_center_policies
- [x] azure_security_center_policy

Padgett:
I'll snag network_watcher(s) and network_security_group(s) for starters

TrevorBramble:
I'll get resource groups and security center polic{y|ies}

Padgett:
Looks like that leaves activity_log_alert(s), I'll do those next

Padgett:
Starting monitor_log_profile(s)

Padgett:
@clintoncwolfe The naming scheme follows Azure's with v1 being `azure` and v2 being `azurerm` which also matches the Terraform providers.

Padgett:
Resolved

PR #89, opened 2018-07-13
Rename azure{,rm}_resource_groups resource

TrevorBramble:
Also added `sort` to the docs list generation because directory contents
are not otherwise ordered deterministically across platforms.

Signed-off-by: Trevor Bramble <tbramble@chef.io>

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -190,7 +190,7 @@ namespace :docs do
   desc 'Prints markdown links for resource doc files to update the README'
   task :resource_links do
     puts "\n"
-    Dir.entries('docs/resources')
+    Dir.entries('docs/resources').sort
        .select { |file| !File.directory?(file) }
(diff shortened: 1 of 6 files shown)


libraries/azure_resource_groups.rb
```diff
@@ -2,27 +2,18 @@
 
 require 'azurerm_resource'
```
Padgett: Should we require the `azurerm_resource_groups` here instead?

Padgett:
Related #81

Rakefile
```diff
@@ -190,7 +190,7 @@ namespace :docs do
   desc 'Prints markdown links for resource doc files to update the README'
   task :resource_links do
     puts "\n"
-    Dir.entries('docs/resources')
+    Dir.entries('docs/resources').sort
```
dmccown: 👍

PR #100, opened 2018-07-25
Introduces rake options task

### LINKED ISSUE #99: No Graph API access prevents terraform from running
# Description of Issue

If you attempt to start a Terraform environment without having permissions to the Graph API your Terraform task will fail with an error "Insufficient Permissions".  We should make Graph an optional configuration item so someone is not blocked if they don't have permissions while developing new resources.

At the moment we have optional features in Terraform:
- msi_vm
- network_watcher

What would be a good interface to allow someone to opt into additional components depending on what they are testing without needing a huge rake command.

If we were to follow the existing pattern we would have something like:
```
$ rake network_watcher graph tf:apply test:integration
```
### Consolidated rake task
```
$ rake with[network_watcher,graph,msi_vm,etc] # modifies .envrc
$ source .envrc # or direnv allow
$ rake tf:apply
$ rake test:integration
$ rake with[] # modifies .envrc to remove all optional components
$ source .envrc # or direnv allow
```
This task could modify your `.envrc` file to add any optional components to your configuration. We could then output something on other rake tasks to output which optional components you are using. As well as noting that tests will be skipped for any optional components you aren't using.

By default all optional components would be disabled. 

### Other ideas

Are there other ideas that would be a better user experience and provide clarity around what components you have enabled/disabled?

dmccown:
### Description

This change introduces a new rake task `options`. You may use this task to set optional components you wish to use. See the README for details on using this feature.

### Issues Resolved

fixes #99 

### Check List

- [x] All tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,38 @@ rake azure:login
 rake azure
 ```
 
+### Development
+
+To run all tests:
+```
+bundle
+rake
+```
+
+To run integration tests:
+```
+bundle
+rake test:integration
+```
+
+To run integration tests including a Network Watcher:
+```
+bundle
+rake network_watcher test:integration
+```
+
+To run a control called `azurerm_virtual_machine`:
+```
+bundle
+rake test:integration[azurerm_virtual_machine]
+```
+
+You may run multiple controls:
+```
+bundle
+rake test:integration[azure_resource_group,azurerm_virtual_machine]
+```
+
 ### Optional Components
 
 By default, rake tasks will only use core components. Optional components have
@@ -280,34 +312,3 @@ direnv allow # or source .envrc
 rake tf:apply
 ```
 
-### Development
-
-To run all tests:
-```
-bundle
-rake
-```
-
-To run integration tests:
-```
-bundle
-rake test:integration
-```
-
-To run integration tests including a Network Watcher:
-```
-bundle
-rake network_watcher test:integration
-```
-
-To run a control called `azurerm_virtual_machine`:
-```
-bundle
-rake test:integration[azurerm_virtual_machine]
-```
-
-You may run multiple controls:
-```
-bundle
-rake test:integration[azure_resource_group,azurerm_virtual_machine]
-```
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -255,3 +245,56 @@ You may run multiple controls:
 bundle
 rake test:integration[azure_resource_group,azurerm_virtual_machine]
 ```
+
+### Optional Components
+
+By default, rake tasks will only use core components. Optional components have
+associated integrations that will be skipped unless you enable these. We have
+the following optional pieces that may be managed with Terraform.
```
dmccown: This doesn't read clearly as the next part I go into is examples and not the optional components.

README.md
```diff
@@ -255,3 +245,56 @@ You may run multiple controls:
...
+
+Optional Components may be combined when running tasks:
+
+```
+rake options[option_1,option_2,option3]
+direnv allow # or source .envrc
+rake tf:apply
+```
+
+To disable optional components just re-run `rake options` without the components you wish to disable:
```
dmccown: Reword:

To disable optional components run `rake options` including only the optional components you wish to enable. Any omitted component will be disabled.

Rakefile
```diff
@@ -206,12 +231,16 @@ namespace :attributes do
   end
 
   task :write_guest_presence_to_file do
-    Dir.chdir(TERRAFORM_DIR) do
-      stdout, stderr, status = Open3.capture3("az ad user list --query=\"length([?userType == 'Guest'])\"")
+    if ENV['GRAPH']
```
dmccown: Switch to check if the key is present. `ENV.key?('GRAPH')`

lib/environment_file.rb
```diff
@@ -0,0 +1,72 @@
...
+    false
+  end
+
+  def replace_key(key, value = nil)
+    new_content = ''
+    found = false
+
+    File.open(@file.path, 'r').each_line do |line|
+      if /export #{key}=.*$/.match?(line)
+        new_content << yield(key, value)
```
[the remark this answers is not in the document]
dmccown: Rubocop! 
```
                        ^
lib/environment_file.rb:79:24: C: Performance/RedundantBlockCall: Use yield instead of formatter
.call.
        new_content << formatter.call(key, value)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^
```

dmccown:
Sourcing `.envrc` from rake doesn't seem to change the environment outside of rake.

PR #120, opened 2018-08-20
Adds Chef/Inspec User-Agent to Azure API calls

r-fennell:
### Description

Adds Chef Open Source UUID as a header on all API calls to Azure, to measure usage/impact.

### Check List

- [x] All tests pass
- [x] `rake lint` passes


davymcaleer:
@r-fennell the User Agent should be:
    INSPEC_USER_AGENT = 'Inspec-18d63047-6cdf-4f34-beed-62f01fc73fc2'
No the pid was actually the ProductID, in this case InSpec

r-fennell:
Should `Chef` removed completely?

libraries/support/azure/rest.rb
```diff
@@ -5,6 +5,10 @@
 
 module Azure
   class Rest
+    USER_AGENT = 'User-Agent'
+    INSPEC_USER_AGENT = 'Chef-Inspec-pid-18d63047-6cdf-4f34-beed-62f01fc73fc2'
+    private_constant  :USER_AGENT, :INSPEC_USER_AGENT
```
dmccown: Strings are already frozen on line 1 https://github.com/inspec/inspec-azure/blob/e11b70eeb1546462f93afc126c27ed32a17b52d1/libraries/support/azure/rest.rb#L1 and these vars should be private by default. I don't think you need line 10.

r-fennell: @dmccown I can't find any documentation indicating default scope of constants is private. I used [this](http://www.chrisrolle.com/en/blog/private-constants-in-ruby) and [this](https://stackoverflow.com/a/11885328) as reference. Is it a new ruby version change?

dmccown: I was wrong in assuming scope, thanks for pointing that out.

libraries/support/azure/rest.rb
```diff
@@ -5,6 +5,9 @@
 
 module Azure
   class Rest
+    USER_AGENT = 'User-Agent'
+    INSPEC_USER_AGENT = 'Chef-Inspec-18d63047-6cdf-4f34-beed-62f01fc73fc2'
```
bmoore-msft: this must be:
INSPEC_USER_AGENT = 'pid-18d63047-6cdf-4f34-beed-62f01fc73fc2'
you can add others if needed

trevorghess: @bmoore-msft "you can add others if needed" so we should generate GUIDs for our specific subproducts rather than tagging them in the agent?

bmoore-msft: We need exactly one string in the "pid-GUID" format...  if there is other info you want to derive from the UA (not related to our tracking) then, you can still add it to the UA, but it will be ignored in the query.  That help?

r-fennell:
@trevorghess Sorry it took so long to get back to this. Let me know if anything else needs changed!

PR #124, opened 2018-08-30
Adds warning when using strings in a where clause

dmccown:
### Description

When we introduced backing resources with structs you could no longer
search plural resources with a string key. This adds backwards
compatability and a deprecation warning to avoid breaking changes.

### Check List

- [x] All tests pass
- [x] `rake lint` passes


libraries/support/azure/deprecations/strings_in_where_clause.rb
```diff
@@ -0,0 +1,26 @@
...
+                "where('#{key}' => '#{value}') to where(#{key}: '#{value}'). " \
+                'Automatic conversion will be removed in version 2.0.'
+
+              transformed[key.to_sym] = value
+            else
+              transformed[key] = value
+            end
+          end
+          old_where(transformed, &block)
+        end
```
TrevorBramble: We can break this into simpler statements for easier comprehension.

Here we emit the deprecation if needed, but always ensure the key is a symbol, whether it was a string or already a symbol.

Also note spelling fix `s/verison/version/` in warning.
```ruby
def where(conditions = {}, &block)
  string_warn = ->(k, v) { warn_deprecation(k, v) if k.is_a?(String); [:"#{k}", v] }

  old_where(conditions.map(&string_warn).to_h, &block)
end

private

def warn_deprecation(key, value)
  warn '[DEPRECATION] String detected in where clause. As of version 1.2 ' \
    'where clauses should use a symbol. Please convert ' \
    "where('#{key}' => '#{value}') to where(#{key}: '#{value}'). " \
    'Automatic conversion will be removed in version 2.0.'
end
```

test/integration/verify/controls/azure_security_center_policies.rb
```diff
@@ -6,7 +6,7 @@
     its('policy_names') { should include('default', resource_group) }
   end
 
-  describe azurerm_security_center_policies.where(name: resource_group) do
+  describe azurerm_security_center_policies.where('name' => resource_group) do
```
TrevorBramble: Why change this?

Do we want one use of the where with a string to make sure it warns and converts? The test should be separate and have a name to say that's the thing we're testing. The above would look like a mistake in the test code to my eyes, possibly evidence that keys _should_ be strings, to others.

libraries/support/azure/deprecations/strings_in_where_clause.rb
dmccown: 👍 That's much easier to read. Thanks!

test/integration/verify/controls/azure_security_center_policies.rb
dmccown: That's left over from my manual test. I need to remove that and replace it with a unit test.

libraries/support/azure/deprecations/strings_in_where_clause.rb
```diff
@@ -0,0 +1,26 @@
+module Deprecations
+  module StringsInWhereClause
+    def self.included(base)
+      base.class_eval do
+        alias_method :old_where, :where
```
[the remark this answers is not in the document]
dmccown: I'm going to go with `filtertable_where`.

PR #127, opened 2018-09-03
Adds SQL Server and SQL Database resources

r-fennell:

### Description

Adds resources for testing of Azure SQL Servers and Databases.

### Check List

- [x] All tests pass
- [x] `rake lint` passes
- [x] All commits have been signed-off for the Developer Certificate of Origin.


libraries/azurerm_sql_servers.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  attr_reader :table
+
+  FilterTable.create
+             .register_column(:ids,        field: :id)
+             .register_column(:names,      field: :name)
+             .register_column(:kinds,      field: :kind)
+             .register_column(:locations,  field: :location)
+             .register_column(:properties, field: :properties)
+             .register_column(:tags,       field: :tags)
```
skpaterson: Should this be:
```
              .register_column(:tags,       field: :tag) 
```

libraries/support/azure/management.rb
```diff
@@ -141,22 +141,38 @@ def virtual_machine_disk(resource_group, id)
...
+                  resource_group: resource_group),
         api_version: '2018-02-01',
       )
     end
 
+    def sql_servers(resource_group)
+      get(
+        url: link(location: 'Microsoft.Sql/servers',
+                  resource_group: resource_group),
+        api_version: '2018-06-01-preview',
```
skpaterson: Might be a good idea to have the `2018-06-01-preview` defined once  and injected here to make it easier to update in the future.  Could come from a configuration file or just a variable.

terraform/azure.tf
```diff
@@ -589,3 +589,20 @@ resource "azurerm_virtual_machine_extension" "virtual_machine_extension" {
     }
 SETTINGS
 }
+
+resource "azurerm_sql_server" "sql-server-1" {
+  name                          = "${var.sql-server-name}"
+  resource_group_name           = "${azurerm_resource_group.rg.name}"
+  location                      = "${var.location}"
+  version                       = "12.0"
```
skpaterson: Any value in this being a variable?  I was bitten recently by a hard-coded OS version going EOL but that's a different story ;)

test/integration/verify/controls/azurerm_sql_servers.rb
```diff
@@ -0,0 +1,9 @@
+resource_group = attribute('resource_group', default: nil)
+
+control 'azurerm_sql_servers' do
+  only_if { ENV['SQL'] }
+
+  describe azurerm_sql_servers(resource_group) do
+    it { should exist }
```
skpaterson: Here would be good to test the plural resource parameters defined above e.g. can check for the `sql_server_name`:
```
its('names'){should include sql_server_name}
```

libraries/azurerm_sql_servers.rb
r-fennell: `tags` is the field name in the singular resource https://github.com/inspec/inspec-azure/blob/198a837941e094e79d9860bab183cc8c73efd408/libraries/azurerm_sql_server.rb#L20 similar to `properties`. Should it still be `tag`?

libraries/support/azure/management.rb
r-fennell: Agreed. Each endpoint has a configured API version and by nature are not consistent. I think there is a future task to consolidate/parameterize the API versions we are using.

terraform/azure.tf
r-fennell: 👍

libraries/azurerm_sql_servers.rb
skpaterson: The use-case I was thinking of is filtering on a single tag e.g. 
```
  describe azurerm_sql_servers(resource_group).where(tag: "my_very_special_tag") do
    it { should exist }
    # you wouldn't need to do the below here but for illustration purposes
    its('tags') { should include "my_very_special_tag" } 
  end
```
The singular resource properties aren't affected, could be named the same or differently.  Might be worth checking if `properties` works as expected because I've only ever seen the plural/singular case.

test/integration/verify/controls/azurerm_sql_servers.rb
r-fennell: 👍

PR #132, opened 2018-09-06
Adds properties documentation

dmccown:
### Description

Adds more advanced documentation on accessing deeper properties as well as referencing official Azure documentation.

### Check List

- [x] All tests pass
- [x] `rake lint` passes


docs/resources/azurerm_ad_user.md.erb
```diff
@@ -10,6 +10,10 @@ an Azure Active Directory user within a Tenant.
 
 <br />
 
+## Azure REST API version
+
+This resource interacts with version `1.6` of the Azure Graph API. For more information see the [official Azure documentation](https://msdn.microsoft.com/Library/Azure/Ad/Graph/api/users-operations#GetAUser).
```
TrevorBramble: Please wrap. =^)

docs/resources/azurerm_ad_user.md.erb
```diff
@@ -164,6 +168,37 @@ The principal name of the user.
...
+```
+
+We may access `provisioningStatus` with:
+
+```
+its('provisionedPlants.first.provisioningStatus') { should eq "Success" }
+```
+
+The api may not always return keys if there is no data associated with that
+key. There may be cases where the deeply nested property may not have the
```
TrevorBramble: For clarity, maybe the first sentence could be reworded more like "The API may not always return keys that do not have any associated data." (Also make sure API is all caps.)

docs/resources/azurerm_ad_users.md.erb
```diff
@@ -10,6 +10,10 @@ some or all Azure Active Directory users within a Tenant.
 
 <br />
 
+## Azure REST API version
+
+This resource interacts with version `1.6` of the Azure Graph API. For more information see the [official Azure documentation](https://msdn.microsoft.com/Library/Azure/Ad/Graph/api/users-operations#GetUsers).
```
TrevorBramble: I guess I expected this section to include something about how the documentation should be referenced with attention to specific resource API versions...

dmccown: As far as I can tell there is no way to select the version of the documentation you want. 😦

docs/resources/azurerm_ad_user.md.erb
TrevorBramble: :point_up:

dmccown: Thanks I missed this one earlier. I do not plan to merge without addressing it.

PR #135, opened 2018-09-10
3 1 ensure storage security

pwelch:
### Description

Add Azure Storage Account Resources

### Check List

- [x] All tests pass
- [x] `rake lint` passes

### DIFF
--- a/libraries/azurerm_storage_account.rb
+++ b/libraries/azurerm_storage_account.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+require 'azurerm_resource'
+
+class AzurermStorageAccout < AzurermSingularResource
+  name 'azurerm_storage_account'
+  desc 'Verifies settings for a Azure Storage Account'
+  example <<-EXAMPLE
+    describe azurerm_storage_account(resource_group: resource_name, name: 'default') do
+      it { should exist }
+      its('secure_transfer_enabled') { should be true }
+    end
+  EXAMPLE
+
+  ATTRS = %i(
+    name
+    id
+    secure_transfer_enabled
+  ).freeze
+
+  attr_reader(*ATTRS)
+
+  def initialize(opt = { name: 'default', resource_group: nil })
+    resp = management.storage_account(opt[:resource_group], opt[:name])
+    return if has_error?(resp)
+
+    @name                    = resp.name
+    @id                      = resp.id
+    @secure_transfer_enabled = resp.properties.supportsHttpsTrafficOnly
+
+    @exists = true
+  end
+
+  def has_secure_transfer_enabled?
+    !!secure_transfer_enabled
+  end
+
+  def to_s
+    "#{name} Storage Account"
+  end
+end
--- a/libraries/azurerm_storage_account.rb
+++ b/libraries/azurerm_storage_account.rb
@@ -20,8 +20,8 @@ class AzurermStorageAccout < AzurermSingularResource
 
   attr_reader(*ATTRS)
 
-  def initialize(opt = { name: 'default', resource_group: nil })
-    resp = management.storage_account(opt[:resource_group], opt[:name])
+  def initialize(name: 'default', resource_group: nil)
+    resp = management.storage_account(resource_group, name)
     return if has_error?(resp)
 
     @name       = resp.name
--- a/libraries/azurerm_storage_account.rb
+++ b/libraries/azurerm_storage_account.rb
@@ -15,7 +15,7 @@ class AzurermStorageAccout < AzurermSingularResource
   ATTRS = %i(
     name
     id
-    secure_transfer_enabled
+    properties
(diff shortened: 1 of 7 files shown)


libraries/support/azure/management.rb
```diff
@@ -94,6 +94,14 @@ def security_center_policies
       )
     end
 
+    def storage_account(resource_group, name)
+      get(
+        url: link(location: 'Microsoft.Storage/storageAccounts',
+                  resource_group: resource_group) + name,
```
r-fennell: Can we use ```"Microsoft.Storage/storageAccounts/#{name}"``` instead? I know there still some examples of concatenation but I think interpolation is best (https://stackoverflow.com/a/10076632)

libraries/azurerm_storage_accounts.rb
```diff
@@ -0,0 +1,32 @@
...
+  attr_reader :table
+
+  def initialize
+    resp = management.storage_accounts
+    return if has_error?(resp)
+
+    @table = resp
+  end
+
+  include Azure::Deprecations::StringsInWhereClause
```
r-fennell: What does this do?

libraries/support/azure/management.rb
dmccown: If were to try to interpolation here it would be something like:
```
url: "#{link(location: ''Microsoft.Storage/storageAccounts',
                  resource_group: resource_group)}#{name}"
```
Would it make more sense to change link to take in a name?
```
        url: link(location: 'Microsoft.Storage/storageAccounts',
                  resource_group: resource_group,
                  name: name),
```

Then we can just assume link always returns a fully crafted URL.

libraries/azurerm_storage_accounts.rb
dmccown: This includes a deprecation warning from previous implementations that keyed plural resources by string keys e.g. `'MyKey'` instead of using symbols. 

By including this you'll monkey patch this resource to redefine the `where` method to check if you used a string instead of a symbol. If you did, then it will convert the string to a symbol and give a deprecation warning.

This should not be included as it's a new resource and you should not be doing any filtering on strings.

libraries/azurerm_storage_account.rb
```diff
@@ -0,0 +1,41 @@
...
+
+require 'azurerm_resource'
+
+class AzurermStorageAccout < AzurermSingularResource
+  name 'azurerm_storage_account'
+  desc 'Verifies settings for a Azure Storage Account'
+  example <<-EXAMPLE
+    describe azurerm_storage_account(resource_group: resource_name, name: 'default') do
+      it { should exist }
+      its('secure_transfer_enabled') { should be true }
```
dmccown: I'd consider making a custom matcher for things that are booleans.

Something like:
```
it { should have_secure_transfer_enabled }
```

libraries/azurerm_storage_accounts.rb
pwelch: An example of poor cargo culting.

libraries/azurerm_storage_account.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  ATTRS = %i(
+    name
+    id
+    secure_transfer_enabled
+  ).freeze
+
+  attr_reader(*ATTRS)
+
+  def initialize(opt = { name: 'default', resource_group: nil })
```
dmccown: I'd use keyword arguments here:

```
   def initialize(name: 'default', resource_group: nil)
```

libraries/support/azure/management.rb
pwelch: Instead of using the `link` method defined in the file?

dmccown: I was proposing changing the the link method to handle a `name` key. 

```
    def link(location:, provider: true, resource_group: nil, name: nil)
      "/subscriptions/#{subscription_id}" \
      "#{"/resourceGroups/#{resource_group}" if resource_group}" \
      "#{'/providers' if provider}" \
      "/#{location}/" \
      "#{name if name}"
end
```
and
```

    def storage_account(resource_group, name)
      get(
        url: link(location: 'Microsoft.Storage/storageAccounts',
                  resource_group: resource_group,
                  name: name),
        api_version: '2017-06-01',
      )
end
```
```

Comment (r-fennell) [reply libraries/support/azure/management.rb]: I'm not a fan of adding any more fields to the `link` method to be honest.
Current params, `subscription`, `resource_group` and `provider` are used in most API calls, but `name` would only be used in at most 50% (singular v. plural).

What is wrong with:
```
    def storage_account(resource_group, name)
      get(
        url: link(location: "Microsoft.Storage/storageAccounts/#{name}",
                  resource_group: resource_group),
        api_version: '2017-06-01',
      )
    end
```

r-fennell:
🚢

PR #136, opened 2018-09-11
Adds Key Vault and SQL Server firewall

r-fennell:
### Description

- Adds Key Vault resource & control.
- Adds Key Vault plural resource & control.
- Adds ability to get SQL Server firewall rules.

### Check List

- [x] All tests pass
- [x] `rake lint` passes


docs/resources/azurerm_key_vaults.md.erb
```diff
@@ -0,0 +1,125 @@
...
+
+## Examples
+
+The following examples show how to use this InSpec audit resource.
+
+### Check Key Vaults are present
+
+    describe azurerm_key_vaults(resource_group: 'my-rg') do
+      it            { should exist }
+      its('names')  { should include 'vault-101' }
```
TrevorBramble: :radioactive: :wink:

test/integration/verify/controls/azurerm_key_vaults.rb
```diff
@@ -0,0 +1,10 @@
+resource_group    = attribute('resource_group', default: nil)
+vault_name        = attribute('key_vault_name', default: nil)
+
+control 'azurerm_key_vaults' do
+
+  describe azurerm_key_vaults(resource_group: resource_group) do
+    it                    { should exist }
+    its('names')          { should include vault_name }
+  end
+end
```
TrevorBramble: There's a lot of unnecessary whitespace in these test files. The process is to write each line as normal, but then push out right-hand segments to match the furthest one.

So if you have some assignments like these:
```ruby
red = [255,0,0]
yellow = [0,255,0]
blue = [0,0,255]
```
then the furthest column is in the `yellow` assignment. Use spaces to push out the assignment operators for `red` and `blue` to match.
```ruby
red    = [255,0,0]
yellow = [0,255,0]
blue   = [0,0,255]
```

Here's what the content of this file should look like.
```ruby
resource_group = attribute('resource_group', default: nil)
￼vault_name     = attribute('key_vault_name', default: nil)
￼
control 'azurerm_key_vaults' do￼
  describe azurerm_key_vaults(resource_group: resource_group) do
    it           { should exist }
    its('names') { should include vault_name }
  end
end
```

docs/resources/azurerm_key_vault.md.erb
```diff
@@ -0,0 +1,109 @@
...
+    depends:
+      inspec-azure:
+        git: https://github.com/inspec/inspec-azure.git
+
+You'll also need to setup your Azure credentials; see the resource pack
+[README](https://github.com/inspec/inspec-azure#inspec-for-azure).
+
+### Version
+
+This resource first became available in 1.1.0 of the inspec-azure resource pack.
```
TrevorBramble: Just marking this line so we don't forget to update it. 1.1.0 was already released, so this will be a later version.

docs/resources/azurerm_key_vaults.md.erb
```diff
@@ -0,0 +1,125 @@
...
+    depends:
+      inspec-azure:
+        git: https://github.com/inspec/inspec-azure.git
+
+You'll also need to setup your Azure credentials; see the resource pack
+[README](https://github.com/inspec/inspec-azure#inspec-for-azure).
+
+### Version
+
+This resource first became available in 1.1.0 of the inspec-azure resource pack.
```
TrevorBramble: Just marking this line so we don't forget to update it. 1.1.0 was already released, so this will be a later version.

docs/resources/azurerm_key_vault.md.erb
```diff
@@ -0,0 +1,109 @@
+---
+title: About the azurerm_key_vault Resource
+platform: azure
+---
+
+# azurerm\_key\_vault
+
+Use the `azurerm_key_vault` InSpec audit resource to test properties and configuration of
+an Azure Key Vault.
+<br />
```
dmccown: I've been adding a link in this section to the REST API documentation in Azure. Depending on when things merge you may want to do the same here. Here's an example of what I've been putting:

```
## Azure REST API version

This resource interacts with version `VERSION` of the Azure Graph API. For more
information see the [official Azure documentation](URL).

At the moment, there doesn't appear to be a way to select the version of the
Azure API docs. If you notice a newer version being referenced in the official
documentation please open an issue or submit a pull request using the updated
version.
```

docs/resources/azurerm_key_vault.md.erb
```diff
@@ -0,0 +1,109 @@
...
+
+<br />
+
+## Parameters
+
+  - `resource_group` - The resource Group to which the Key Vault belongs.
+  - `vault_name` - The unique name of the Key Vault.
+
+
+## Properties
```
dmccown: I renamed `Properties` to `Attributes` since there's also a field called `properties`.

docs/resources/azurerm_key_vault.md.erb
```diff
@@ -0,0 +1,109 @@
...
+
+- `id`
+- `name`
+- `location`
+- `type`
+- `tags`
+- `properties`
+    
+For more please see the [Microsoft Azure Documentation](https://docs.microsoft.com/en-us/rest/api/keyvault/vaults/get)
```
dmccown: Let's move this to the top so it's consistent with the rest of the documentation. I also have a block about accessing undocumented attributes in my PR. We should copy that over.

https://github.com/inspec/inspec-azure/pull/132/files#diff-b1337c86da30376d088558c672ee2dfbR128

dmccown: I've been putting that section at the end of the attributes and before `Matchers`.

docs/resources/azurerm_key_vaults.md.erb
```diff
@@ -0,0 +1,125 @@
+---
+title: About the `azurerm_key_vaults` Resource
+platform: azure
+---
+
+# azurerm\_key\_vaults
+
+Use the `azurerm_key_vaults` InSpec audit resource to test properties and configuration of Azure Key Vaults.
+<br />
```
dmccown: Same here with the version information and API document link.

docs/resources/azurerm_key_vault.md.erb
r-fennell: Good call. I can Add this to the SQL resources too.

test/integration/verify/controls/azurerm_key_vaults.rb
r-fennell: Yep - an oversight!

PR #140, opened 2018-09-17
Exempts tags from being turned into a struct

### LINKED ISSUE #138: How to properly validate tags
### 🎛 Description

I have defined a simple control to verify the existence of required tags with an Azure resource:

```
control 'default' do
  describe azurerm_virtual_network(resource_group: 'my-rg', name: 'my-nw') do
    tags = {
      'My-Tag-1' => 'foo',
      'My-Tag-2' => 'bar',
      'My-Tag-3' => 'baz'
    }

    it { should exist }
    its('tags') { should eq tags }
  end
end
```

Using the following inspec.yml profile:

```
name: default
depends:
- name: inspec-azure
  url: https://github.com/inspec/inspec-azure/archive/1.1.0.tar.gz
supports:
- platform: azure
```

When I run the tests, I get the following error:

```
  ×  default: 'my-nw' Virtual Network (1 failed)
     ✔  'my-nw' Virtual Network should exist
     ×  'my-nw' Virtual Network tags should eq {"My-Tag-1"=>"foo", "My-Tag-2"=>"bar", "My-Tag-3"=>"baz"}

     expected: {"My-Tag-1"=>"foo", "My-Tag-2"=>"bar", "My-Tag-3"=>"baz"}
          got: #<struct :"My-Tag-1"="foo", :"My-Tag-2"="bar", :"My-Tag-3"="baz">

     (compared using ==)

     Diff:
     @@ -1,5 +1,2 @@
     -"My-Tag-1" => "foo",
     -"My-Tag-2" => "bar",
     -"My-Tag-3" => "baz",
     +"#<struct :\"My-Tag-1\"=\"foo\", :\"My-Tag-\"=\"var\", :\"My-Tag-3\"=\"baz\">"
```

According to the documentation, my approach should be correct. How would I need to change the test to make it work?

### 🌍 InSpec and Platform Version

```
bundle exec inspec version
2.2.78
```

dmccown:
Turning tags into a struct makes it difficult to compare the expected
sets of tags with the actual set of tags returned. There's also the
potential for dashes in tag names. In these cases the struct methods
are only accessable by calling `send` on the struct with the key names.

This change preserves tags as a hash.

fixes #138

### Description

Please describe what this change achieves. Ensure you have read the [Contributing to InSpec](https://github.com/inspec/inspec-azure/CONTRIBUTING.MD) document before submitting.

### Issues Resolved

### Check List

- [ ] All tests pass
- [ ] `rake lint` passes
- [ ] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

### DIFF
--- a/libraries/support/azure/service.rb
+++ b/libraries/support/azure/service.rb
@@ -14,6 +14,11 @@ module Azure
       end
     end
 
+    EXEMPTED_TAGS = %i(
+      tags
+    ).freeze
+    private_constant :EXEMPTED_TAGS
+
     def with_cache(cache)
       @cache = cache
     end
@@ -42,7 +47,11 @@ module Azure
       return data unless data.is_a?(Hash)
       return data if data.empty?
 
-      ResponseStruct.create(data.keys.map(&:to_sym), data.values.map { |v| to_struct(v) })
+      exempted = data.slice(*EXEMPTED_TAGS)
+      data.reject! { |k, _| EXEMPTED_TAGS.include?(k.to_sym) }
+
+      ResponseStruct.create(data.keys.map(&:to_sym) + exempted.keys.map(&:to_sym),
+                            data.values.map { |v| to_struct(v) } + exempted.values)
     end
 
     private
--- a/test/unit/support/azure/service_test.rb
+++ b/test/unit/support/azure/service_test.rb
@@ -3,8 +3,10 @@ require_relative '../../../../libraries/support/azure/service'
 
 describe Azure::Service do
   let(:address)         { { street: '123 Fake St.', city: 'Nowhere', state: 'NO' } }
+  let(:tags)            { { 'tag-1' => 'my-custom-tag', 'tag-2' => 'my-other-tag' } }
   let(:john)            { { name: 'John', age: 20, address: address } }
   let(:jane)            { { name: 'Jane', age: 25, address: {} } }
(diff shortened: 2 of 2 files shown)


libraries/support/azure/service.rb
```diff
@@ -14,6 +14,11 @@ def fetch(key)
       end
     end
 
+    EXEMPTED_TAGS = %i(
```
r-fennell: `EXEMPTED_ATTRIBUTES` / `EXEMPTED_FIELDS`? In case we find other hashes we want to keep, which arent called `tags`?

libraries/support/azure/service.rb
```diff
@@ -42,13 +47,22 @@ def to_struct(data)
...
+
+      ResponseStruct.create(data.keys.map(&:to_sym) + exempted.keys.map(&:to_sym),
+                            data.values.map { |v| to_struct(v) } + exempted.values)
     end
 
     private
 
     attr_reader :required_attrs
 
+    def slice!(data, keys)
```
r-fennell: could we make it clearer by changing `keys` -> `to_exclude` or something similar?

dmccown: How about `keys_to_select`?

r-fennell: 👍

PR #142, opened 2018-09-19
Adds KeyVault Keys & Vault Client

r-fennell:
### Description

- Adds `azurerm_key_vault_key` resource / control
- Adds `azurerm_key_vault_keys` resource / control
- Adds new Vault client for interrogating Azure `https://{vault-name}.vault.azure.net` APIs

### Check List

- [x] All tests pass
- [x] `rake lint` passes


libraries/azurerm_key_vault_secrets.rb
```diff
@@ -0,0 +1,35 @@
...
+  EXAMPLE
+
+  attr_reader :table
+
+  FilterTable.create
+             .register_column(:ids,          field: :id)
+             .register_column(:attributes,   field: :attributes)
+             .register_column(:contentTypes, field: :contentType) { |table, _value| table[:contentType] || nil }
+             .register_column(:managed,      field: :managed)     { |table, _value| table[:managed] || false }
+             .register_column(:tags,         field: :tags)        { |table, _value| table[:tags] || nil }
```
TrevorBramble: These lines get pretty long so you could just use `_` instead of `_value`, since we don't get anything from it anyway.

On the "or" values, `nil` is already returned when the key doesn't exist, so those could be removed unless it's important to communicate that `nil` is a possible value here. In which case I'd change all of these to use `fetch` (`table.fetch(:contentType, nil)`, etc.) including the one that should return `false` on a miss.

libraries/azurerm_resource.rb
```diff
@@ -20,6 +20,11 @@ def graph
                 .for_tenant(tenant_id)
   end
 
+  def vault(vault_name)
+    Azure::Vault.new
+                .with_client(vault_client(vault_name))
```
TrevorBramble: One line is fine here too. =^)

libraries/support/azure/service.rb
```diff
@@ -69,21 +69,25 @@ def set_reader(name, value, override)
...
     end
 
-    def get(url:, api_version:, error_handler: nil, unwrap: nil)
+    def get(use_cache = true, url:, api_version:, error_handler: nil, unwrap: nil)
       confirm_configured!
 
-      cache.fetch(url) do
+      if use_cache
+        body = cache.fetch(url)
+      end
```
TrevorBramble: This could be turned into a conditional clause. `body = cache.fetch(url) if use_cache`

libraries/support/azure/service.rb
```diff
@@ -69,21 +69,25 @@ def set_reader(name, value, override)
       self
     end
 
-    def get(url:, api_version:, error_handler: nil, unwrap: nil)
+    def get(use_cache = true, url:, api_version:, error_handler: nil, unwrap: nil)
```
TrevorBramble: There's no need to add a positional argument for this, and I'd prefer you didn't.
```ruby
def get(url:, api_version:, error_handler: nil, unwrap: nil, use_cache: true)
```
```ruby
get(url: '/keys', api_version: '2016-10-01', use_cache: false)
```

terraform/azure.tf
```diff
@@ -103,6 +103,7 @@ resource "azurerm_key_vault" "disk_vault" {
       "delete",
       "get",
       "set",
+      "list",
```
TrevorBramble: Alphabetical, please. =^)

test/integration/verify/controls/azurerm_key_vault_key.rb
```diff
@@ -0,0 +1,12 @@
+vault_name = attribute('key_vault_name', default: nil)
+key_name   = attribute('key_vault_key_name', default: nil)
+
+control 'azurerm_key_vault_key' do
+
```
TrevorBramble: Please remove the empty line after opening each `control` block.

libraries/azurerm_key_vault_key.rb
```diff
@@ -0,0 +1,53 @@
...
+  private
+
+  VALID_VERSION_REGEX = Regexp.new('^([0-9a-f]{8})([0-9a-f]{4})([0-9a-f]{4})([0-9a-f]{2})([0-9a-f]{2})([0-9a-f]{12})$')
+
+  def valid_version?(version)
+    version.downcase
+           .scan(VALID_VERSION_REGEX)
+           .any?
+  end
+end
```
TrevorBramble: Pretty much everything I said about `AzurermKeyVaultSecret` applies to this file too. =^)

docs/resources/azurerm_key_vault_secret.md.erb
```diff
@@ -0,0 +1,140 @@
...
+### value
+The secret value.
+
+### attributes
+The secret management attributes.
+
+### kid
+If this is a secret backing a KV certificate, then this field specifies the corresponding key backing the KV certificate.
+
+### content_type
```
dmccown: Let's escape `_` in headers to be safe.

docs/resources/azurerm_key_vault_secret.md.erb
```diff
@@ -0,0 +1,140 @@
...
+## Attributes
+
+  -  `id`
+  -  `value`
+  -  `attributes`
+  -  `kid`
+  -  `content_type`
+  -  `managed`
+  -  `tags`
+
```
dmccown: Extra blank line.

docs/resources/azurerm_key_vault_secret.md.erb
```diff
@@ -0,0 +1,140 @@
...
+### content_type
+The content type of the secret.
+
+### managed
+True if the secret's lifetime is managed by key vault. If this is a secret backing a certificate, then managed will be true.
+
+### tags
+Resource tags applied to the Secret.
+
+
```
dmccown: Extra blank line.

docs/resources/azurerm_key_vault_secrets.md
```diff
@@ -0,0 +1,121 @@
...
+
+### managed
+
+Filters the results to include only those Secrets which are not managed by the Vault. This is a boolean value.
+
+    describe azurerm_key_vault_secrets('my-vault').where{ managed.eql?(false) } do
+      it { should_not exist }
+    end
+
+## Attributes
```
dmccown: Attributes duplicated.

docs/resources/azurerm_key_vault_secrets.md
```diff
@@ -0,0 +1,121 @@
...
+
+### contentType
+Type of the secret value such as a password.
+
+### managed
+True if the secret's lifetime is managed by key vault. If this is a key backing a certificate, then managed will be true.
+
+### tags
+Resource tags applied to the Key.
+
```
dmccown: Extra blank line.

libraries/support/azure/vault.rb
```diff
@@ -0,0 +1,77 @@
...
+    def initialize
+      @required_attrs = %i(rest_client)
+      @page_link_name = 'nextLink'
+    end
+
+    def keys
+      get(
+        url: '/keys',
+        api_version: '2016-10-01',
+        use_cache: false,
```
dmccown: Why don't we want to use the cache for keys and secrets?

libraries/support/azure/vault.rb
```diff
@@ -0,0 +1,77 @@
...
+      )
+    end
+
+    def secret(secret_name, secret_version)
+      get(
+        url: "/secrets/#{secret_name}/#{secret_version}",
+        api_version: '2016-10-01',
+        unwrap: unwrap,
```
dmccown: Why do we need `unwrap` here? For graph we are using unwrap since things are paginated.

libraries/support/azure/vault.rb
r-fennell: Caching had to be disabled in two places:
1. Because of how caching is performed in https://github.com/inspec/train/blob/master/lib/train/transports/azure.rb I found that requesting secrets from both `https://red.vault.azure.net` and `https://blue.vault.azure.net` would always return the secrets cached from whichever called first. Instead, we disable caching via `    inspec.backend.disable_cache(:api_call)` when we ask for a Vault client and always instantiate a new one, using the correct vault name.
2. The same problem was present in https://github.com/inspec/inspec-azure/blob/Add-Keys/libraries/support/azure/service.rb#L75 - if you request the `/secrets` or `/keys` endpoint more than once, even from different Vault clients, the result of the first call would always be returned.

libraries/support/azure/vault.rb
r-fennell: The default unwrap (https://github.com/inspec/inspec-azure/blob/Add-Keys/libraries/support/azure/service.rb#L86) looks for an attribute `value` in the response, which with Azure usually contains the JSON representation of what we want to map to struct. 
Unfortunately for Keys, there is an actual attribute called `value`, and so instead of mapping the entire response, it mapped only this `value` field to struct.

libraries/support/azure/vault.rb
dmccown: Sounds like a good work around at the moment. We should drop an issue to revisit caching later.

libraries/support/azure/vault.rb
dmccown: Thanks!

TrevorBramble:
@r-fennell Have your TODO items in the PR description been handled?

docs/resources/azurerm_key_vault_key.md.erb
```diff
@@ -0,0 +1,128 @@
...
+### Other Attributes
+
+There are additional attributes that may be accessed that we have not
+documented. Please take a look at the [Azure documentation](##-Azure-REST-API-version).
+Any attribute in the response may be accessed with the key names separated by
+dots (`.`).
+
+The API may not always return keys that do not have any associated data. There
+may be cases where the deeply nested property may not have the desired
+attribute along your call chain. If find yourself writing tests against
```
TrevorBramble: "If [you] find"

Sorry, I've corrected it in the other files just now. Could you amend this one as well?

docs/resources/azurerm_key_vault_secret.md.erb
```diff
@@ -0,0 +1,137 @@
...
+### Other Attributes
+
+There are additional attributes that may be accessed that we have not
+documented. Please take a look at the [Azure documentation](##-Azure-REST-API-version).
+Any attribute in the response may be accessed with the key names separated by
+dots (`.`).
+
+The API may not always return keys that do not have any associated data. There
+may be cases where the deeply nested property may not have the desired
+attribute along your call chain. If find yourself writing tests against
```
TrevorBramble: "If [you] find"

r-fennell:
> @r-fennell Have your TODO items in the PR description been handled?

@TrevorBramble Train PR has been merged but awaiting the latest release to be promoted, which will allow the dependancy bump. Hopefully it will be done today!

PR #143, opened 2018-10-03
Scopes storage_accounts resource by resource group

dmccown:
### Description

When trying to iterate over all resource groups and storage accounts we
currently can only ask for all storage accounts. This means we try to
look up every storage account for every resource group.

This fix allows one to scope their storage accounts query by resource
group.

Removes rake target we no longer need.

### Check List

- [x] All tests pass
- [x] `rake lint` passes

### DIFF
--- a/libraries/support/azure/management.rb
+++ b/libraries/support/azure/management.rb
@@ -102,9 +102,10 @@ module Azure
       )
     end
 
-    def storage_accounts
+    def storage_accounts(resource_group)
       get(
-        url: link(location: 'Microsoft.Storage/storageAccounts'),
+        url: link(location: 'Microsoft.Storage/storageAccounts',
+                  resource_group: resource_group),
         api_version: '2017-06-01',
       )
     end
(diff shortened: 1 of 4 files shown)


libraries/support/azure/management.rb
```diff
@@ -102,9 +102,10 @@ def storage_account(resource_group, name)
       )
     end
 
-    def storage_accounts
+    def storage_accounts(resource_group)
       get(
-        url: link(location: 'Microsoft.Storage/storageAccounts'),
+        url: link(location: 'Microsoft.Storage/storageAccounts',
+                  resource_group: resource_group),
```
TrevorBramble: So this is OK even when `resource_group` is `nil`? Are you sure it shouldn't be conditionally added based on whether it has a value?

dmccown: We check for `nil` here: https://github.com/inspec/inspec-azure/blob/master/libraries/support/azure/management.rb#L286

I also have a test that does the query with and without a `resource_group`. https://github.com/inspec/inspec-azure/pull/143/files#diff-5496d1f35534e24f5840421d4cad3a57

pwelch:
Good catch! 👍

PR #149, opened 2018-10-17
Added azurerm_aks_cluster resource

jmassardo:
Signed-off-by: James Massardo <jmassardo@chef.io>

### Description
This PR adds new resources to provide basic auditing for Azure AKS clusters

### DIFF
--- a/terraform/azure.tf
+++ b/terraform/azure.tf
@@ -634,3 +634,29 @@ resource "azurerm_sql_database" "sql-database" {
   depends_on          = ["azurerm_sql_server.sql-server"]
   tags {}
 }
+
+resource "azurerm_kubernetes_cluster" "test" {
+  name                = "inspecakstest"
+  location            = "${azurerm_resource_group.rg.location}"
+  resource_group_name = "${azurerm_resource_group.rg.name}"
+  dns_prefix          = "inspecaksagent1"
+
+  agent_pool_profile {
+    name            = "default"
+    count           = 5
+    vm_size         = "Standard_D1_v2"
+    os_type         = "Linux"
+    os_disk_size_gb = 30
+  }
+  linux_profile {
+    admin_username = "inspecuser1"
+
+    ssh_key {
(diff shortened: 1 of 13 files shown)


terraform/azure.tf
```diff
@@ -620,3 +620,29 @@ resource "azurerm_sql_database" "sql-database" {
...
+    count           = 5
+    vm_size         = "Standard_D1_v2"
+    os_type         = "Linux"
+    os_disk_size_gb = 30
+  }
+  linux_profile {
+    admin_username = "inspecuser1"
+
+    ssh_key {
+      key_data = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDDQTKriRSJjQOurKaXSQKb5rmkv1SBtBKIvy4K8+dlUp4xot+CMcJVb9UmTl9EuVuwcktTt/9iEmFVkDIovbaiYWkT/LVeJHQkwJQ2ezykG3eipvAPeQX9RcUTTRP69DV/01f3nVQZOM7HJl/xswuX/ps8eBEcov9hzdEKDq5c0KPokb9NHNrToax5t+fd2xC41hkQL6obYwD/gDw4t1tGQV2Xvq5ellZhUnLQfSQ5cjec/FOMCKtwAcOKjQqZh2BXqv2MQaxCfjwX7gmhKNABUrq7+BQQgJA1YxS69lhFozzhTmH88PHRL3V/Plfm6ismIU+UdjF49iTnJ69fj2Vv"
```
pwelch: What is this key from? 

I think we should extract this out into a terraform variable.

jmassardo: That is basically a dummy key as the terraform resource requires a key before it will provision the systems.

I can move it to a variable. Are there any other changes needed?

terraform/variables.tf
```diff
@@ -2,6 +2,7 @@ variable "subscription_id" {}
 variable "client_id" {}
 variable "client_secret" {}
 variable "tenant_id" {}
+variable "ssh_key" {}
```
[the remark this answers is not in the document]
TrevorBramble: What is the `SSH_KEY` value expected to be? How is it used? These need to be documented, so someone (me) who doesn't know what it's about can do whatever is needed in order to run the integration tests, or use it in the field.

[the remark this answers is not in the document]
jmassardo: I added a note to the readme about the SSH_KEY

jmassardo:
I updated the files. Please let me know if there are any other changes needed.

jmassardo:
I missed some changes for the plural resource so I fixed them and amended the PR.

terraform/azure.tf
```diff
@@ -620,3 +620,29 @@ resource "azurerm_sql_database" "sql-database" {
   depends_on          = ["azurerm_sql_server.sql-server"]
   tags {}
 }
+
+resource "azurerm_kubernetes_cluster" "test" {
+  name                = "inspecaks"
```
dmccown:
```suggestion
  name                = "inspecakstest"
```

The `agent_pool_profile` name should be unique from the `azurerm_kubernetes_cluster` name. When they are the same tests were failing.

PR #152, opened 2018-10-18
Add ? to boolean method

r-fennell:
Add proper naming to method.

### DIFF
--- a/libraries/azurerm_storage_account.rb
+++ b/libraries/azurerm_storage_account.rb
@@ -30,7 +30,7 @@ class AzurermStorageAccout < AzurermSingularResource
     @exists = true
   end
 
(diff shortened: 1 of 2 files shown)


libraries/azurerm_storage_account.rb
```diff
@@ -30,7 +30,7 @@ def initialize(name: 'default', resource_group: nil)
     @exists = true
   end
 
-  def have_recently_generated_access_key
+  def has_recently_generated_access_key?
```
pwelch: Good catch!

r-fennell: Was kind of shoe horned after the fact but yes 😅

Issue #144, opened 2018-10-04
undefined method `networkSecurityGroup' for #<#<Class:0x00000004337e18>:0x00000004336298>


rdtechie:
### 🎛 Description

**🐛 bug report**

When I'm using the azurerm_subnet resource and check for NSG presence, I'm getting a error message when there's no NSG attached:

`undefined method 'networkSecurityGroup' for #<#<Class:0x00000004337e18>:0x00000004336298>`

The complete output is as following:
```
[DEPRECATED] use a specific azure resources instead of 'azure_generic_resource'. See https://github.com/inspec/inspec/issues/3131

Profile: InSpec Profile (azure-compliance-profile)
Version: 0.1.0
Target:  azure://af9f0308-1806-444b-9d9e-e10bda0f9aa5

  ✔  check-if-resource-group-exists: Ensure Resource Group Exists
     ✔  Resource Groups with name == "TestNetwork" should exist
  ✔  check-resource-group-properties: Ensure Resource Group properties are correct
     ✔  azure_generic_resource should have tags
     ✔  azure_generic_resource location should eq "northeurope"
     ✔  azure_generic_resource tag_count should equal 2
  ×  azurerm_subnet: 'TestSubnet' subnet (1 failed)
     ✔  'TestSubnet' subnet should exist
     ✔  'TestSubnet' subnet name should eq "TestSubnet"
     ✔  'TestSubnet' subnet type should eq "Microsoft.Network/virtualNetworks/subnets"
     ✔  'TestSubnet' subnet address_prefix should eq "10.10.10.0/24"
     ×  'TestSubnet' subnet nsg
     undefined method `networkSecurityGroup' for #<#<Class:0x00000002a88458>:0x00000002a83b60>


Profile: Azure Resource Pack (inspec-azure)
Version: 1.2.0
Target:  azure://af9f0308-1806-444b-9d9e-e10bda0f9aa5

     No tests executed.

Profile Summary: 2 successful controls, 1 control failure, 0 controls skipped
Test Summary: 8 successful, 1 failure, 0 skipped
```
When I attach the NSG to the subnet, the test completes succesful:
```
[DEPRECATED] use a specific azure resources instead of 'azure_generic_resource'. See https://github.com/inspec/inspec/issues/3131

Profile: InSpec Profile (azure-compliance-profile)
Version: 0.1.0
Target:  azure://af9f0308-1806-444b-9d9e-e10bda0f9aa5

[...]

PR #155, opened 2018-11-02
Add graceful-ish handling of missing response data

TrevorBramble:
### Description

This is more scar tissue around Azure APIs not adhering to a response contract such that the key for a given item is not returned when there is no attached item of that type. (If a VM has no attached drives, it does not return `{... "drives": [] ...}`, it just does not include that element at all.)

When testing against our struct data with dot-notation instead of hash index references (`data.foo` vs. `data[:foo]`) a `NoMethodError` exception would be raised if `foo` wasn't included in the API response.

This change handles that scenario by returning `nil` when there is a reference to an unkown key. It's not ideal, but it'll help while we get more feedback from our current model and decide whether a larger change is appropriate.

### Check List

- [X] New functionality includes tests
- [X] All tests pass
- [X] `rake lint` passes
- [X] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

### DIFF
--- a/libraries/support/azure/response_struct.rb
+++ b/libraries/support/azure/response_struct.rb
@@ -7,8 +7,13 @@ module Azure
         def key?(key)
           members.include?(key)
         end
-
         alias_method :keys, :members
+
+        private
+
+        def method_missing(_name)
+          nil
+        end
       end.new(*values)
     end
   end
--- a/test/unit/support/azure/response_struct_test.rb
+++ b/test/unit/support/azure/response_struct_test.rb
@@ -1,0 +1,43 @@
+require_relative '../../test_helper'
+require_relative '../../../../libraries/support/azure/response_struct'
+
+describe Azure::ResponseStruct do
+  let(:keys) { %i(foo bar baz) }
+  let(:values) { %w{foo bar baz} }
+  let(:data) { keys.zip(values).to_h }
+
+  subject { Azure::ResponseStruct.create(keys, values) }
+
+  it 'should produce a Struct' do
+    expect(subject).must_be_kind_of Struct
+  end
+
+  it 'should respond to given keys' do
+    keys.map { |key| expect(subject).must_respond_to key }
+  end
+
+  it 'should have expected values for given keys' do
+    data.map { |key, value| expect(subject.send(key)).must_equal value }
+  end
+
+  it 'should return nil for unknown keys' do
+    expect(subject.unknown_key).must_be_nil
+  end
+
+  describe 'should support some Hash methods' do
+    it 'should support .key?()' do
+      expect(subject).must_respond_to :key?
+
+      expect(subject.key?(keys.first)).must_equal true
+      expect(subject.key?(:unknown_key)).must_equal false
+    end
+
+    it 'should support .keys()' do
+      expect(subject).must_respond_to :keys
+
+      expect(subject.keys).must_equal keys
+
+      expect(subject.keys).must_equal subject.members
+    end
+  end
+end


dmccown:
Do you have thoughts on how we could handle deep hash navigation more safely?

```
describe azurerm_my_resource do
  its('properties.key1.key2') { should eq 'foo' }
end
```

In that case if `key1` doesn't exist can we safely return an object that is `nil` and responds to `key2`?

dmccown:
fixes #144

TrevorBramble:
> Do you have thoughts on how we could handle deep hash navigation more safely?
> 
> ```
> describe azurerm_my_resource do
>   its('properties.key1.key2') { should eq 'foo' }
> end
> ```
> In that case if `key1` doesn't exist can we safely return an object that is `nil` and responds to `key2`?

@dmccown Don't forget we convert the response recursively, so all member hashes are converted to structs as well. https://github.com/inspec/inspec-azure/blob/05b6a619a963730fd7fa01b378bb3c86b3ab3c51/libraries/support/azure/service.rb#L45-L54

dmccown:
In the case I'm referencing above the first key doesn't exist. So `properties.key1` returns `nil`. Your call then fails with:
```
        1: from (irb):1
NoMethodError (undefined method `key2' for nil:NilClass)
```

TrevorBramble:
@dmccown I follow you now. Null object?

r-fennell:
+1 for a NullObject to handle deep navigation. It also opens the possibility to return a meaningful error message instead of `nil`, but i'm not sure how this would play out with controls (may give false passes).

dmccown:
Yeah I think a NullObject is appropriate here. Returning a `'nil` seems appropriate since the Azure REST API isn't giving us an answer with a 200 response code. 

We do have a gap on handling non 200 responses (e.g. 4xx and 5xx). In those cases we should think about the right kind of error to give so someone will know the API gave an unexpected response.

libraries/azurerm_virtual_machines.rb
```diff
@@ -66,7 +66,7 @@ def with_data_disks
       disks = Array(vm.properties.storageProfile.dataDisks)
       disks = disks.select { |disk| disk.key?(:managedDisk) }
 
-      Azure::ResponseStruct.create(vm.members << :data_disks, vm.values << disks.collect(&:name))
+      Azure::Response.create(vm.members << :data_disks, vm.values << disks.collect(&:name))
```
dmccown: This name reads much better.

test/integration/verify/controls/azurerm_virtual_machine.rb
```diff
@@ -1,35 +1,39 @@
...
-data_disk_names       = attribute('windows_vm_data_disks', default: nil)
-monitoring_agent_name = attribute('monitoring_agent_name', default: nil)
+resource_group            = attribute('resource_group',        default: nil)
+win_name                  = attribute('windows_vm_name',       default: nil)
+win_id                    = attribute('windows_vm_id',         default: nil)
+win_location              = attribute('windows_vm_location',   default: nil)
+win_tags                  = attribute('windows_vm_tags',       default: nil)
+win_os_disk_name          = attribute('windows_vm_os_disk',    default: nil)
+win_data_disk_names       = attribute('windows_vm_data_disks', default: nil)
+win_monitoring_agent_name = attribute('monitoring_agent_name', default: nil)
```
dmccown: It just dawned on me that we may have a dangerous pattern here. If we fail to import an attribute:
```
win_name                  = attribute('windows_vm_name',       default: nil)
```
e.g. `win_name` comes back as `nil`. Then the test:
```
 its('name')                       { should eq win_name }
```
`name` would return `nil` and assert that it's the same as `win_name`. Using this pattern would every test pass if you forgot to pass in the attribute file now?

TrevorBramble: Well the test run bombs in various ways when I knock chunks out of the attributes file, or omit it from the InSpec invocation, but yeah it's all just accidental breakage, not failing tests.

I'm not sure this is a problem. If it is, what do you think a solution might be? Just adding more examples like `its('name') { should_not be_nil }` or `its('name') { should_not be_empty }`? Wouldn't it be needed for all of the attributes, then?

dmccown: You're right the test will still return the actual value if it's there. Even if we fed in that we expected a `nil` value we'd fail because the actual Azure calls would return a value. 

I'll resolve this conversation.

PR #162, opened 2019-01-25
Tidy Rakefile a little

r-fennell:

### Description
Tidying up Rakefile

### Issues Resolved
Removing distinct exit code from integration test runs, and enabling lint on the same.

### Check List

- [x] All tests pass
- [x] `rake lint` passes

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -18,15 +18,13 @@ REQUIRED_ENVS = %w{AZURE_CLIENT_ID AZURE_CLIENT_SECRET AZURE_TENANT_ID}.freeze
 
 task default: :test
 desc 'Testing tasks'
-task test: ['test:unit', 'test:integration']
-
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -72,7 +70,7 @@ namespace :inspec do
     puts stdout
 
     %w{errors}.each do |type|
-      abort("InSpec check failed with syntax #{type}!") if !!(/[1-9]\d* #{type}/ =~ stdout)
+      abort("InSpec check failed with syntax #{type}!") if /[1-9]\d* #{type}/ =~ stdout
```
dmccown: `!!` was added to explicitly turn the results from `=~` into a boolean. With this change you are relying on the truthiness of the return value which would either be `0` or a `#<MatchData...>`.

PR #188, opened 2019-07-17
Update README.md

mjingle:
Signed-off-by: Mary Jinglewski <mjinglewski@chef.io>

### Description

Updating the versioning reference to reflect best practices.

### Check List

- [ ] All tests pass
- [ ] `rake lint` passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,13 +58,15 @@ Example inspec.yml:
 name: my-profile
 title: My own Oneview profile
 version: 0.1.0
-inspec_version: '>= 2.2.7'
+inspec_version: '>= 4.6.9'
 depends:
   - name: inspec-azure
-    url: https://github.com/inspec/inspec-azure/archive/master.tar.gz
+    url: https://github.com/inspec/inspec-azure/archive/x.tar.gz
 supports:
   - platform: azure
 ```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -58,14 +58,16 @@ Example inspec.yml:
 name: my-profile
 title: My own Oneview profile
```
rmoles: Small nitpick. Would it be possible to update this to Azure profile?

README.md
```diff
@@ -58,14 +58,16 @@ Example inspec.yml:
...
 version: 0.1.0
-inspec_version: '>= 2.2.7'
+inspec_version: '>= 4.6.9'
 depends:
   - name: inspec-azure
-    url: https://github.com/inspec/inspec-azure/archive/master.tar.gz
+    url: https://github.com/inspec/inspec-azure/archive/x.tar.gz
```
zenspider: same as before:

> We should figure out what our format is going to be and template this a bit better (eg `inspec-X.Y.tar.gz` or `compat` instead of `inspec`... arbitrary).... but just NOT saying "master" right now is more important than getting the format right. I'd even go so far as to ask that this get published immediately.

README.md
zenspider: Yes please. (what's Oneview? and why is MS so bad at naming things?)

PR #215, opened 2019-10-30
Bump Inspec, Adds Azurerm Disks Plural resource

r-fennell:

### Description
- Make use of latest Inspec via `inspec-bin` gem.
- Adds `azurerm_virtual_machine_disks` plural resource w/ tests, docs, etc.
- Adds Azure ID `to_h` conversion to base `azurerm_resource` class to remove growing number of 'magic number' string manipulations (see [example](https://github.com/inspec/inspec-azure/pull/215/files#r340765374)).

### Check List

- [x] All tests pass
- [x] `rake lint` passes


libraries/azurerm_resource.rb
```diff
@@ -44,6 +45,34 @@ def assign_fields_with_map(map, struct)
...
+    begin
+      split = azure_id.split('/').reject!(&:empty?)
+      raise unless !split.empty? && split.length.even?
+      hash = {}
+
+      i = 0
+      counter = 0
+      while counter < split.length/2
+        # Remove plural from API noun.
+        k = split[i].end_with?('s') ? split[i][0..-2] : split[i]
```
skpaterson: Are there any cases here where the API resource noun legitimately ends in an 's' e.g. `log_analytics` maybe?  If so I'm wondering if it might be preferable to leave this field unaltered and go with exactly what appears in the fully qualified url.

docs/resource_creation_guide.md
```diff
@@ -77,11 +77,11 @@ Consider the following:
 - Update variables and Terraform outputs as needed.
 
 ```
-resource_group = attribute('resource_group',  default: nil)
-subnet         = attribute('subnet_name',     default: nil)
-tags           = attribute('subnet_tags',     default: nil)
-id             = attribute('subnet_id',       default: nil)
-location       = attribute('subnet_location', default: nil)
+resource_group = attribute('resource_group',  value: nil)
```
skpaterson: If making the `default` -> `value` change might be best also to update `attribute` to `input` at the same time.

libraries/azurerm_resource.rb
r-fennell: Change added 👍

docs/resource_creation_guide.md
r-fennell: Done!

PR #218, opened 2019-11-06
Upgrade Terraform to v0.12

r-fennell:
### Description

Updates integration tests to use terraform 0.12.

### Check List

- [x] All tests pass
- [x] `rake lint` passes


terraform/azure.tf
```diff
@@ -386,39 +386,39 @@ resource "null_resource" "set_log_profile_retention" {
...
+  resource_group_name   = azurerm_resource_group.rg.name
+  network_interface_ids = [azurerm_network_interface.nic2[0].id]
   vm_size               = "Standard_DS2_v2"
 
-  tags {
+  tags = {
     Description = "Externally facing Linux machine with SSH access"
   }
 
-  identity = {
```
rmoles: Should this `=` have been removed? The trend in this PR seems to be adding `=` rather than removing.

terraform/azure.tf
```diff
@@ -524,43 +523,43 @@ resource "azurerm_kubernetes_cluster" "cluster" {
...
-  resource_group_name = "${azurerm_resource_group.rg.name}"
-  app_service_plan_id = "${azurerm_app_service_plan.app_service_plan.id}"
+  location            = azurerm_resource_group.rg.location
+  resource_group_name = azurerm_resource_group.rg.name
+  app_service_plan_id = azurerm_app_service_plan.app_service_plan.id
   https_only          = "true"
 
-  identity = {
-    type              = "SystemAssigned"
+  identity {
```
rmoles: Same query as above.

terraform/azure.tf
r-fennell: Its a change in how the provider handles the arguments (block vs hash). Not sure why things have changed but here we are 🤷‍♂ .

PR #222, opened 2019-11-21
Adds ability to check Network Watcher flow logs.

r-fennell:


libraries/azurerm_network_watcher.rb
```diff
@@ -22,20 +22,27 @@ class AzurermNetworkWatcher < AzurermSingularResource
...
+    @resource_group = resource_group
     @exists = true
   end
 
   def to_s
     "'#{name}' Network Watcher"
   end
 
+  def flow_logs
+    return nil if @nsg.nil?
```
rmoles: This could be changed to `!!@nsg`. Have seen elsewhere that this preferred over `.nil?`

PR #285, opened 2020-08-18
Add apply_only rake task

rmoles:
This will aid in retrying resource creation in CI
Signed-off-by: Ross Moles <rmoles@chef.io>


rmoles:
> LGTM - might be nice to add a description for the task also

I deliberately didn't add a description for this task as it shouldn't be needed outside of CI

PR #286, opened 2020-08-27
Re-write more resources with the new backend

omerdemirok:

### Description

README
- Update with the new resources

OLD DOCS
- Add depreciation message

NEW DOCS
- Add docs for the new resources
- Fix some typos in the previously created ones

BACKEND
- Add `to_s` method to `AzureResourceProbe`
- Delete `to_s` method of `NullResponse` to prevent possible confusion
- Handle resource aggregation in `AzureGenericResources` and simplify plural static resources

### Issues Resolved

N/A

### Check List

- [x] All tests pass
- [x] `rake lint` passes


docs/resources/azure_aks_cluster.md
```diff
@@ -0,0 +1,93 @@
...
+```
+## Parameters
+
+| Name                           | Description                                                                       |
+|--------------------------------|-----------------------------------------------------------------------------------|
+| resource_group                 | Azure resource group that the targeted resource resides in. `MyResourceGroup`     |
+| name                           | Name of the AKS cluster to test. `ClusterName`                                      |
+| resource_id                    | The unique resource ID. `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/providers/Microsoft.ContainerService/managedClusters/{ClusterName}` |
+
+Either one of the parameter sets can be provided for a valid query:
```
rmoles, line 43: Based on this, am I right to assume that this is also valid?
```
describe azure_aks_cluster(resource_id: 'example') do
   it { should exist }
 end
```
If so, it would be good to provide this an example also.

test/unit/resources/azure_aks_cluster_test.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'helper'
```
rmoles, line 1: Can you double-check that this file is valid in your local IDE. I'm not seeing any syntax highlighting in the GitHub UI.

docs/resources/azurerm_virtual_network.md
```diff
@@ -1,4 +1,4 @@
----
+r---
```
rmoles, line 1: This looks like an error

test/unit/resources/azure_aks_cluster_test.rb
omerdemirok: Yes, it is a valid file: https://github.com/inspec/inspec-azure/blob/master/test/unit/resources/helper.rb

libraries/azure_backend.rb
```diff
@@ -673,6 +671,10 @@ def method_missing(method_name, *args, &block)
   def respond_to_missing?(*several_variants)
     super
   end
+
+  def to_s
+    "Property is missing! The following are available: #{item.keys.map(&:to_s)}"
```
rmoles, line 676: It would be nice to re-word this slightly...
`Property is missing! The following properties are available: #{item.keys.map(&:to_s)}`

libraries/azure_event_hub_event_hub.rb
```diff
@@ -0,0 +1,44 @@
+require 'azure_generic_resource'
+
+class AzureEventHubEventHub < AzureGenericResource
+  name 'azure_event_hub_event_hub'
+  desc 'Verifies settings for Event Hub Event Hub'
```
rmoles, line 5: `Event Hub` is repeated twice here and in the desc for the backward compatible resource

rmoles (approved):
Thanks, @omerdemirok. Looks good to me 💯

PR #295, opened 2020-09-11
Add azure_postgresql_server(s) resources

omerdemirok:

### Description

Resources:
- Add fully backward compatible `azure_postgresql_server(s)` resources

Tests:
- Update `only_if` conditions

### Issues Resolved

N/A

### Check List

- [x] All tests pass
- [x] `rake lint` passes


rmoles (approved):
Thanks @omerdemirok

libraries/azure_postgresql_server.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  private
+
+  def create_configurations
+    return unless exists?
+    resource_uri = id + '/configurations'
+    query = {
+      resource_uri: resource_uri,
+    }
+    confs = get_resource(query)[:value].map { |c| [c[:name], c] }.to_h
```
rmoles, line 43: Is there a scenario where `[:value]` could be empty?

libraries/azure_postgresql_server.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  # Resource specific methods can be created.
+  # `return unless exists?` is necessary to prevent any unforeseen Ruby error.
+  # Following methods are created to provide the same functionality with the current resource pack >>>>
+  # @see https://github.com/inspec/inspec-azure
+
+  private
+
+  def create_configurations
```
rmoles, line 37: I find `create_configurations` slightly misleading. Perhaps `get_configurations`.

test/integration/verify/controls/azurerm_postgresql_servers.rb
```diff
@@ -2,7 +2,7 @@
 postgresql_server_name = input('postgresql_server_name', value: nil)
 
 control 'azurerm_postgresql_servers' do
-  only_if { ENV['SQL'] }
+  only_if { !postgresql_server_name.nil? }
```
rmoles, line 5: There probably should be a corresponding update [here](https://github.com/chef-partners/inspec-azure-ci/blob/ef34e0826a90be63dbe31843b7eb9db16319df7a/Jenkinsfile#L111)
Weird that this was here when there was no logic surrounding the creation of the resources.

libraries/azure_postgresql_server.rb
omerdemirok: Good catch!
Probably configurations are default and they are not empty, but, I think it is better to check before using them.

PR #300, opened 2020-09-21
Enable users to define api versions for the additional property API calls

omerdemirok:

### Description

Users can provide a specific API version for the additional properties in the supported resources.

### Issues Resolved

This will fix the failing CIS benchmark tests caused by incompatible API versions.

### Check List

- [x] All tests pass
- [x] `rake lint` passes


libraries/azure_backend.rb
```diff
@@ -690,7 +688,7 @@ def respond_to_missing?(*several_variants)
   end
 
   def to_s
-    "Property is missing! The following properties are available: #{item.keys.map(&:to_s)}"
+    "#{type}/#{name} has following properties: #{item.keys.map(&:to_s)}."
```
rmoles, line 691: This should probably be: `has the following`

omerdemirok: Updated.

rmoles (approved):
Thanks @omerdemirok

PR #309, opened 2020-09-28
Add virtual_disk(s) resources

omerdemirok:

### Description

- Add `azure_virtual_disk(s)` resources

### Issues Resolved

N/A

### Check List

- [x] All tests pass
- [x] `rake lint` passes


docs/resources/azure_virtual_machine_disk.md
```diff
@@ -0,0 +1,117 @@
...
+
+Either one of the parameter sets can be provided for a valid query:
+- `resource_id`
+- `resource_group` and `name`
+
+## Properties
+
+| Property          | Description |
+|---------------------------|-------------|
+| encryption_enabled        | Indicates whether the encryption is enabled or not. Note that this will return `nil` unless the encryption status is defined on the resource explicitly. |
```
rmoles, line 55: I wonder if we could add a resource method or attribute reader, for this property and default it `false` or whatever the value should be for encryption is disabled.

rmoles: In fact, I see that you have already created a method for this. It would be good to just default the value to `false` or `disabled`

test/integration/verify/controls/azurerm_virtual_machine_disk.rb
```diff
@@ -19,7 +19,7 @@
   end
 
   describe azurerm_virtual_machine_disk(resource_group: resource_group, name: unencrypted_disk_name) do
-    its('encryption_enabled') { should be false }
```
rmoles, line 22: perhaps the name field could be updated to: `name: attached_disk_name`

docs/resources/azure_virtual_machine_disk.md
omerdemirok: This is created for backward compatibility The property used for making this conclusion, in the old API, is being presented through multiple properties with the latest API version. 
Considering this behavior change over time, I thought it is safer to return the value only if it is provided explicitly from the endpoint. (`properties.encryptionSettingsCollection.enabled`)

Even though returning `encryption_enabled = false` is fail-safe when it is encrypted, I thought it should be the user's responsibility to make this call by returning `encryption_enabled = nil`.

rmoles: Returning nil in this scenrio will actually break backward compatibility. The old `azurerm` resource did this
```
!!properties.encryptionSettings.enabled
```
As an arbitrary example, below is what would be returned...
```
2.7.0 :001 > foo = true
2.7.0 :002 > !!foo
 => true
2.7.0 :003 > foo = false
2.7.0 :004 > !!foo
 => false
2.7.0 :005 > foo = nil
2.7.0 :006 > !!foo
 => false
```

As you can see, in the old resource this would have actually returned `false` when `nil` was returned

omerdemirok: This does not take into account the [properties.encryption](https://docs.microsoft.com/en-us/rest/api/compute/disks/get#encryption) property where it shows: `Encryption property can be used to encrypt data at rest with customer managed keys or platform managed keys.`. (Available in the latest API.)

We can check this value and eventually return `false` if it doesn`t exist either. But, leaving this decision to the user still seems more accurate than making an assumption. The new resources default to the latest API version and this makes it difficult to return the same result across the different versions.

test/integration/verify/controls/azurerm_virtual_machine_disk.rb
omerdemirok: This is be fixed now.

rmoles (approved):
Thanks @omerdemirok

PR #310, opened 2020-09-29
Remove trailing space from monitor_log scopes

omerdemirok:

### Description

- Remove trailing space from the `azure_monitor_activity_log_alert#scopes`
- Update an example 
- Update `azure_aks_cluster#enabled_logging_types` and `azure_aks_cluster#disabled_logging_types` methods to take multiple diagnostic settings into account

### Issues Resolved

Fixes failing controls for `Activity Log Alert` in the CIS Benchmark tests

### Check List

- [x] All tests pass
- [x] `rake lint` passes


libraries/azure_monitor_activity_log_alert.rb
```diff
@@ -32,7 +32,7 @@ def operations
 
   def scopes
     return unless exists?
-    properties&.scopes
+    properties&.scopes&.map { |scope| scope.delete_suffix('/') }
```
rmoles, line 35: Debating, whether it would be useful to have this after `super(opts, true)`
```
properties&.scopes&.map { |scope| scope.delete_suffix('/') }
```
This could be helpful for anyone that may access `scopes` via the full path rather than the method. If it was to be moved it may require a `!` to be added to update properties in place.

omerdemirok: Good catch!

omerdemirok: This is fixed now.

rmoles (approved):
Thanks @omerdemirok

PR #316, opened 2020-10-02
Add azure_lock(s) resources

omerdemirok:

### Description

- Add azure_lock(s) resources

### Issues Resolved

N/A

### Check List

- [x] All tests pass
- [x] `rake lint` passes


libraries/azure_backend.rb
```diff
@@ -236,7 +236,7 @@ def get_resource(opts = {})
                                 allow: %i(api_version),
                                 opts: opts)
     api_version = opts[:api_version] || 'latest'
-    if opts[:resource_uri].include?('providers')
+    if opts[:resource_uri].scan('providers').size == 1
```
rmoles, line 239: would `.match?` be sufficient here if we are only needing one instance of `providers`?
```
if opts[:resource_uri].match?('providers')
```

docs/resources/azure_locks.md
```diff
@@ -0,0 +1,114 @@
...
+
+This InSpec audit resource has the following special matchers. For a full list of available matchers, please visit our [Universal Matchers page](https://www.inspec.io/docs/reference/matchers/).
+
+### exists
+
+The control will pass if the filter returns at least one result. Use `should_not` if you expect zero matches.
+```ruby
+describe azure_locks(resource_group: 'rg-1', resource_name: 'my-VM', resource_type: 'Microsoft.Compute/virtualMachines') do
+  it { should exist }
+end
```
rmoles, line 110: This functionality would probably be more applicable on the singular resource. Usually, on the plural resource, we would provide a way to filter on the resources, rather than just pull back a specific resource. Perhaps there's a reason for this though?

omerdemirok: Management locks can be defined at the resource level. If this is the case, the resource identifier has to be provided. And, because there are multiple locks on a resource, the plural inspec resource is used. (+ backward compatibility)
https://docs.microsoft.com/en-us/rest/api/resources/managementlocks/listatresourcelevel

rmoles: Makes sense to include this for backward compatibility

rmoles (approved):
Thanks @omerdemirok

PR #322, opened 2020-10-07
Adding subscription level check for diagnostic settings

### LINKED ISSUE #321: Adding diagnostic settings check at subscription level
## Detailed Description

Adding a diagnostic settings check at the Azure subscription level.

## Context

This will allow users to verify that the correct diagnostic settings have been enabled across their subscription.

## Possible Implementation

About to create a PR to add this functionality as well as adding the relevant docs.

voojoe:
### Description

Adding in a number of new attributes for the azure_subscription resource in order to allow checks for configuration of the diagnostic settings.

### Issues Resolved

#321 

### Check List

- [✓] All tests pass
- [✓] `rake lint` passes
- [✓] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>

### DIFF
--- a/libraries/azure_subscription.rb
+++ b/libraries/azure_subscription.rb
@@ -94,7 +94,7 @@ class AzureSubscription < AzureGenericResource
       result.push setting.name
     end
     result
-  end 
+  end
 
   def diagnostic_settings_locations
     return nil if diagnostic_settings.first.nil?
@@ -112,7 +112,7 @@ class AzureSubscription < AzureGenericResource
       result.push setting.properties.eventHubName
     end
     result
-  end 
+  end
 
   def diagnostic_settings_enabled_logging_types
     return nil if diagnostic_settings.first.nil?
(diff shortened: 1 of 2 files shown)


libraries/azure_subscription.rb
```diff
@@ -76,6 +76,62 @@ def logical_locations
...
+      result += setting.properties.logs.select(&:enabled).map(&:category)
+    end
+    result
+  end
+
+  def diagnostic_settings_disabled_logging_types
+    return nil if diagnostic_settings.first.nil?
+    result = []
+    diagnostic_settings.each do |setting|
+      result += setting.properties.logs.reject(&:enabled).map(&:category)
```
rmoles, line 130: it's perhaps worthwhile to use the ruby safe navigator in each of these methods, in case there is a scenario where one of the values does not exist.
```
result += setting.properties&.logs&.reject(&:enabled)&.map(&:category)
```

voojoe: Hey @rmoles thanks for the suggestion, I've pushed that now

rmoles (approved):
Thanks for the PR @voojoe

PR #324, opened 2020-10-20
rewrite storage account to new generic resource way (WIP)

srb3:
Signed-off-by: srb3 <sbrown@chef.io>

### Description

converts the storage account resources to the azure generic resource
This is WIP as it is currently blocked by at least https://github.com/inspec/inspec-azure/pull/323

PR #327, opened 2020-11-03
Update backend to support accepting headers at resource level

omerdemirok:

### Description

- Update backend to support accepting HTTP request headers at the resource level
- Add a helper module for hash keys modification recursively

### Issues Resolved

Unblocks #324 

### Check List

- [x] All tests pass
- [x] `rake lint` passes

### DIFF
--- a/libraries/backend/azure_connection.rb
+++ b/libraries/backend/azure_connection.rb
@@ -24,6 +24,9 @@ class AzureConnection
   # @return [String] the graph api endpoint url
   attr_reader :graph_api_endpoint_url
 
+  # @return [String] the endpoint suffix for storage accounts
+  attr_reader :storage_endpoint_suffix
+
   # @return [String] the graph api endpoint api version, e.g. v1.0
   attr_reader :graph_api_endpoint_api_version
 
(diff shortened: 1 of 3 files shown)


libraries/backend/azure_connection.rb
```diff
@@ -157,7 +162,10 @@ def fail_api_query(resp, message = nil)
     message ||= "Unsuccessful HTTP request to Azure REST API.\n"
     message += "HTTP #{resp.status}.\n"
     body = resp.body
-    unless body.empty?
+    if body.empty?
```
rmoles, line 165: Is this change intentional? I would have imagined the message body can have content when successful also.

omerdemirok: This is in the `fail_api_query` method which is called only when the HTTP status is not 200.
It is just a refactoring to prevent increasing the conditional statement depth. The functionality is the same.

rmoles: cool no worries

rmoles (approved):
Looks good to me

Issue #357, opened 2021-01-04
allow_port_from_internet of azure_network_security_group always return true when ICMP is allowed


qiantaosap:
# Describe the problem

When an Azure network security group has rule allowing ICMP, **allow_port_from_internet** of Azure resource **azure_network_security_group** will always return true.

## Possible Solution

Need to add protocol check inside allow_port_from_internet. (In Azure NSG settings, port number is set to * by default for ICMP rules)

PR #358, opened 2021-01-05
Fix allow port from internet

srb3:
### Description
When an ICMP rule is created in Azure the port number is set to *. If ICMP is allowed from the internet then the `allow_port_from_internet?` method returns true for all ports tested, this is not the desired behaviour - this fix makes the port check criteria only consider rules that are `not_icmp`  

### Issues Resolved
https://github.com/inspec/inspec-azure/issues/357

### Check List

- [ ] All tests pass
- [ ] `rake lint` passes
- [ ] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>


TheLunaticScripter (approved):
LGTM!

PR #385, opened 2021-06-01
Support policy insights query result(s)

sathish-progress:
### Description
This PR allows the Azure Policy Insights Query results to be queryable 

### Check List

- [X] New functionality includes tests
- [X] All tests pass
- [X] `rake lint` passes
- [X] All commits have been signed-off for the Developer Certificate of Origin. See <https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco>


sathish-progress:
Hey @trickyearlobe, Please find the PR for policy insights

libraries/azure_generic_resources.rb
```diff
@@ -113,7 +113,9 @@ def api_version_used_for_query_state
   def self.populate_filter_table(raw_data, table_scheme)
     filter_table = FilterTable.create
     table_scheme.each do |col_field|
-      filter_table.register_column(col_field[:column], field: col_field[:field])
```
trickyearlobe, line 116: Hi @sathish-progress , could you explain what this change is for?

sathish-progress: the filter-table library allows us to produce flattened and uniq arrays as columns, as [described](https://github.com/inspec/inspec/blob/master/dev-docs/filtertable-usage.md#a-colors-method-that-will-return-a-flattened-and-uniqd-array-of-values)

and here for some reason, the `style` option was not allowed to be set at individual resources.. Hence the change..

libraries/azure_policy_insights_query_result.rb
```diff
@@ -0,0 +1,98 @@
...
+    describe azure_policy_insights_query_result(policy_definition: '', resource_id: '') do
+      it { should exist }
+    end
+  EXAMPLE
+
+  def initialize(opts = {})
+    raise ArgumentError, 'Parameters must be provided in an Hash object.' unless opts.is_a?(Hash)
+    super(opts)
+    @resource_provider = specific_resource_constraint('Microsoft.Authorization/policyDefinitions', @opts)
+    validate_parameters(required: %i(policy_definition resource_id))
```
trickyearlobe, line 16: Hi @sathish-progress ,
While I think it's OK for the MVP, I'm not convinced we should force the provision of `policy_definitition` and `resource_id` parameters to do filtering on the API side. There are a much broader set of filters we might want to apply, and we should therefore make each one optional and build the filter dynamically.

The API can return a full set of results if we omit the API side filters, which we can then process using the filter table `.where{}` clause. That would make it quite easy to do the caching stuff we discussed on the demo.

Thoughts?

sathish-progress: We are doing that already in [plural resource](libraries/azure_policy_insights_query_results.rb) and here are some [examples](test/integration/verify/controls/azure_policy_insights_query_results.rb) 

This is a singular resource to send more detailed feedback to automate for individual resources..

libraries/azure_policy_insights_query_result.rb
```diff
@@ -0,0 +1,98 @@
...
+  end
+
+  def failed_resource?
+    @failed_resource ||= false
+  end
+end
+
+# Provide the same functionality under the old resource name.
+# This is for backward compatibility.
+class AzurermPolicyInsightsQueryResult < AzurePolicyInsightsQueryResult
```
trickyearlobe, line 80: Hi @sathish-progress 
Don't think we need this bit of code... looks like you forgot to remove it from the cut/paste of another resource that has been renamed. All it does is duplicate the new resource under it's older name and add a deprecation warning.

There wasn't an old `AzurermPolicyInsightsQueryResult` that anyone was using.

libraries/azure_policy_insights_query_result.rb
trickyearlobe: Ah yeah... sorry, my bad... hadn't noticed this was the singular resource.

libraries/azure_policy_insights_query_result.rb
sathish-progress: Okay yeah that completely makes sense.. previously, I was under the impression if someone is referring to all the resources in `Azurerm*` format .. I will remove that ..

libraries/azure_policy_insights_query_results.rb
```diff
@@ -0,0 +1,77 @@
...
+    opts[:resource_uri] = ['providers',  resource_provider, opts.delete(:policy_definition),
+                           'providers/Microsoft.PolicyInsights/policyStates/latest/queryResults'].join('/')
+    opts[:add_subscription_id] = true
+    opts[:method] = 'post'
+
+    super
+
+    return if failed_resource?
+
+    table_schema = [
```
trickyearlobe, line 30: Hi @sathish-progress 
It might be worth exposing all the fields we get from Azure. An example of missing fields that are actually quite useful:-

- complianceReasonCode (sometimes populated with a failure type)
- policyAssignmentParameters (these are sometimes overridden from default in sub scopes)
- policyDefinitionAction (which tells us things like whether this is a DINE or AINE policy)
- timestamp (which in my MVP resource I convert to a Ruby Time class for more useful filtering)

sathish-progress: yeah make sense, I will add those

trickyearlobe (commented):
In general this looks quite good for an MVP.
I'd be obliged if someone could run the tests as I am unable to right now.

If the tests are OK, let's get it merged once @ppradhan9 is happy with the documentation parts and then we can iterate on it.

trickyearlobe:
Hi @sathish-progress , while making a wrapper profile for a demo, I noticed it's necessary to provide a `policy_definition` parameter for the plural resource which requires us to know in advance which policy we wish to query.

In my previous MVP, I didn't need that (and hence I can get results for ALL policies at once if required). The endpoint I'm using is `providers/Microsoft.PolicyInsights/policyStates/latest/queryResults`

Could you give it a tweak?

I also noticed that when we get a lot of results, the backend doesn't paginate properly and only retrieves the first thousand results. I'll raise a seperate PR to fix that.

trickyearlobe:
Oh... forgot to mention... pulling all results at once and caching them in a variable makes it pretty fast to run lots of different aggregations.

IanMadd (approved):

libraries/azure_policy_insights_query_result.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+  EXAMPLE
+
+  def initialize(opts = {})
+    raise ArgumentError, 'Parameters must be provided in an Hash object.' unless opts.is_a?(Hash)
+    super(opts)
+    @resource_provider = specific_resource_constraint('Microsoft.Authorization/policyDefinitions', @opts)
+    validate_parameters(required: %i(policy_definition resource_id))
+    @azure_resource_id = @opts.delete(:resource_id)
+    @policy_definition = @opts.delete(:policy_definition)
```
Nirbhay1997, line 18: Is this because of line 16 ?

sathish-progress: no, we need it for line 57, where we check the resource URI and it  internally forces us to not use any other parameter even though it is valid for resource

libraries/azure_policy_insights_query_result.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  private
+
+  attr_reader :azure_resource_id, :policy_definition, :resource_provider
+
+  def build_query_params
+    query_params = { resource_uri: @opts[:resource_uri] }
+    query_params[:query_parameters] = { '$filter' => "resourceId eq '#{azure_resource_id}'" }
+    # Use the latest api_version unless provided.
```
Nirbhay1997, line 50: Just for my understanding how can we provide api_version explicitly?

Nirbhay1997 (approved):

libraries/azure_policy_insights_query_result.rb
sathish-progress: something like the below..
```
describe azure_policy_insights_query_result(policy_definition: '', resource_id: '', api_version: '2020-08-01') do
      it { should exist }
    end
```

PR #395, opened 2021-06-14
adding azure key vault extension

Rohit1509:
Signed-off-by: Rohit <rohit.prasad.joshi@sap.com>

### Description

Please describe what this change achieves. Ensure you have read the [Contributing to InSpec](https://github.com/inspec/inspec-azure/CONTRIBUTING.MD) document before submitting.

### Issues Resolved

### Check List

- [x] All tests pass
- [x] `rake lint` passes


libraries/azure_key_vault.rb
```diff
@@ -88,6 +88,15 @@ def diagnostic_settings
       },
     )
   end
+
+  def diagnostic_settings_logs
+    return nil if diagnostic_settings.first.nil?
+    result = []
+    diagnostic_settings.each do |setting|
+      result += setting.properties&.logs&.select(&:category)&.map(&:enabled)
```
sathish-progress, line 96: chained safe check operation is an [anti-pattern](https://www.alchemists.io/articles/ruby_antipatterns/#_safe_navigation_operator).. 

also this allows to add `nil` object to result array which is going to raise a TypeError..

libraries/azure_key_vault.rb
```diff
@@ -88,6 +88,15 @@ def diagnostic_settings
       },
     )
   end
+
+  def diagnostic_settings_logs
+    return nil if diagnostic_settings.first.nil?
```
sathish-progress, line 93:
```suggestion
    return nil if diagnostic_settings.nil? || diagnostic_settings.empty?
```

Rohit1509: Thanks for the suggestion. Fixed this!

libraries/azure_key_vault.rb
Rohit1509: it was done similarly for 'azure_subscription' resource. If this is not the appropriate way could you suggest a change here ? I want to check the value for 'properties.logs.enabled'. 
Thanks!

Rohit1509: In normal scenario, I tried running a control against this and it works.

sathish-progress: yeah this will break if in case properties doesn't have an attribute logs! or for that matter any other attribute in the given chain

sathish-progress: here are some of the suggestions, choose which you prefer,

```suggestion
      logs = setting.properties&.logs
      next unless logs
      
      result +=  logs.reduce([]){|arr, log| arr.push(log.enabled) if log.category; arr}
```

there is a less efficient approach but with better readability as well for the last line 
```
logs.map{|log| log.enabled if log.category }.compact
```

Rohit1509: fixed. Thanks!

sathish-progress (approved):
LGTM

PR #564, opened 2021-11-19
Removed broken type conversion in NSG resource

voojoe:
Signed-off-by: Joe McCrea <joe.mccrea@sap.com>

### Description

We discovered an issue when scanning certain Network Security Groups.

Within the 'destination_port_ranges' function, which takes 'properties' as an argument, a type conversion is done on 'properties' to convert it to a hash. This does not work and causes a nil value to be assigned to the new 'properties_hash' variable.

Any time this is then used it does not return the expected value, meaning that the proper checks are not carried out in order to return the correct value for the destination port ranges. The only reason this is not very visible or obvious is due to the values returned from the Azure API, as when 'destinationPortRanges' is returned, 'destinationPortRange' is not returned at all in the response. However when 'destinationPortRange' is returned, an empty value is returned for 'destinationPortRanges' here, which makes it look like the checks have worked even though they have not.

### Issues Resolved

Bug in specific situations when the Network Security Group resource is used.

### Check List

- [x] All tests pass
- [x] `rake lint` passes

### DIFF
--- a/libraries/azure_network_security_group.rb
+++ b/libraries/azure_network_security_group.rb
@@ -170,10 +170,9 @@ class AzureNetworkSecurityGroup < AzureGenericResource
   end
 
   def destination_port_ranges(properties)
-    properties_hash = properties.to_h
-    return Array(properties.destinationPortRange) unless properties_hash.include?(:destinationPortRanges)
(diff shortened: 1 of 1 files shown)


sathish-progress (approved):
LGTM

PR #644, opened 2022-03-03
Adding Logging Enabled Check to Azure Key Vault

jnikles1:
Signed-off-by: Justin Nikles <justin.nikles@sap.com>

### Description

Adds a check specifically for the AuditEvent log setting, which is set to true when logging is enabled https://docs.microsoft.com/en-us/azure/key-vault/general/howto-logging?tabs=azure-cli#enable-logging

### Issues Resolved

### Check List

- [x] All tests pass
- [x] `rake lint` passes


jnikles1:
the diagnostic_settings_logs method doesn't sufficiently report whether logging is enabled or not. If any setting in the log block that is returned is not enabled (whether it signals that logging is enabled or not) the array reports a false flag. For this reason, the additional method was added to specifically check the field which signals that logging is enabled.

libraries/azure_key_vault.rb
```diff
@@ -99,6 +99,18 @@ def diagnostic_settings_logs
     end
     result
   end
+
+  def has_logging_enabled?
+    return false if diagnostic_settings.nil? || diagnostic_settings.empty?
+    diagnostic_settings.each do |setting|
+      logs = setting.properties&.logs
+      next unless logs
+      logs.each do |log|
```
sathish-progress, line 108: this block will return `[]` when `log.category` is not an `AuditEvent`

libraries/azure_key_vault.rb
```diff
@@ -99,6 +99,18 @@ def diagnostic_settings_logs
...
+        next unless log.category == 'AuditEvent'
+        return log.enabled
+      end
+    end
```
sathish-progress, line 112:
```suggestion
    diagnostic_settings.each do |setting|
      log = setting.properties&.logs&.detect{|log| log.category == 'AuditEvent' }
      return log&.enabled
    end
```

jnikles1:
hi @sathish-progress - thanks for the suggestion. It's now implemented as suggested

SathishAchilles (approved):
LGTM

sathish-progress (approved):

PR #652, opened 2022-04-28
Bugfix for Azure NSG Internet Ingress RegEx

jnikles1:
Signed-off-by: Justin Nikles <justin.nikles@sap.com>

### Description

We received reports on false positives where Azure NSGs were flagged as allowing internet ingress, but no such rule was allowing ingress for the flagged ports from the internet. After some investigation, it was found that there was a rule being flagged that allowed traffic for the private address space 10.0.0.0. Due to the RegEx in the source_open method, any rule on an IP address ending in 0.0.0.0 would be flagged, IE 10.0.0.0, 20.0.0.0, etc. This PR fixes this bug and results in proper flagging of the internet IP CIDR.

### Issues Resolved

no issue was opened - but improper flagging of non-internet CIDR ranges was fixed

### Check List

- [ ] All tests pass
- [x] `rake lint` passes

### DIFF
--- a/libraries/azure_network_security_group.rb
+++ b/libraries/azure_network_security_group.rb
@@ -197,7 +197,7 @@ class AzureNetworkSecurityGroup < AzureGenericResource
   def source_open?(properties)
     properties_hash = properties.to_h
     if properties_hash.include?(:sourceAddressPrefix)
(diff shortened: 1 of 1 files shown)


sathish-progress:
> @sathish-progress - do you know if your buildkite is broken? It seems to be blocked

I have run the job.. it should be fine now

sathish-progress (approved):

